PR #26, opened 2020-03-11
CAMEL-12191: Camel-Kafka: create an example on Kubernetes/Openshift by using Strimzi

mabulgu:
Camel-Strimzi example is added due to the issue [CAMEL-12191](https://issues.apache.org/jira/browse/CAMEL-12191). 

Any comments/reviews will be appreciated.

p.s. Another example will be added to Camel Examples as well without the spring-boot runtime. I started this example with spring-boot before the camel-spring-boot repository is separated.


examples/camel-example-spring-boot-strimzi/src/main/java/org/apache/camel/example/strimzi/StrimziRouteBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class StrimziRouteBuilder extends RouteBuilder {
+
+    @Override
+    public void configure() throws Exception {
+
+        from("timer://foo?period=1000")
+        .routeId("Producer Route")
+        .setBody(simple("Hi, this is Camel-Strimzi example from {{environment}} environment"))
+        .to("kafka:{{producer.topic}}?brokers={{kafka.bootstrap.url}}&sslEndpointAlgorithm={{kafka.sslEndpointAlgorithm}}&securityProtocol={{kafka.securityProtocol}}&sslTruststoreLocation={{kafka.truststore.location}}&sslTruststorePassword={{kafka.truststore.password}}");
```
davsclaus: Instead of all these uri options, then most of them can be configured in application.properties with camel.component.kafka.xxx=yyy

examples/camel-example-spring-boot-strimzi/src/main/java/org/apache/camel/example/strimzi/StrimziRouteBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+        ;
+
+        from("kafka:{{producer.topic}}?brokers={{kafka.bootstrap.url}}&sslEndpointAlgorithm={{kafka.sslEndpointAlgorithm}}&securityProtocol={{kafka.securityProtocol}}&sslTruststoreLocation={{kafka.truststore.location}}&sslTruststorePassword={{kafka.truststore.password}}")
+        .routeId("Consumer Route")
+        .log("${body}")
+        ;
```
davsclaus: Move semi colon up

examples/camel-example-spring-boot-strimzi/src/main/resources/spring/camel-context.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<beans xmlns="http://www.springframework.org/schema/beans" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="        http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd        http://camel.apache.org/schema/spring       http://camel.apache.org/schema/spring/camel-spring.xsd">
+    <bean class="org.apache.camel.example.strimzi.StrimziRouteBuilder" id="strimziRouteBuilder" />
+    <camelContext id="camel" xmlns="http://camel.apache.org/schema/spring">
```
davsclaus: You do not need a spring xml file. Just add route builder class and have spring annotation on class, eg @ component. See the other examples

mabulgu: I saw a couple of examples like this: setting the camel context in the xml file and create the bean here. Regarding to your comment i suppose we have to change those as well. (I.e. kafka avro example)

examples/camel-example-spring-boot-strimzi/src/main/java/org/apache/camel/example/strimzi/StrimziRouteBuilder.java
mabulgu: Is that a common approach to set the uris in the .properties file? Isn't that sth out-of-property? Should we consider those as just configurable strings -from properties file- in a plain java class?

davsclaus: Setting the brokers option is likely better on component level so its set for all endpoints. The producer.topic etc is fine as-is etc.

examples/camel-example-spring-boot-strimzi/src/main/resources/spring/camel-context.xml
davsclaus: Yeah those examples should be updated

davsclaus: spring boot is not spring xml anymore. Only a few dedicated examples that shows you can still do camel xml routes.

mabulgu: i know, but one also can use xmls. i just wasn't sure what to do since in this case there may be some camel best practices which i might not master the concept of. So we have to follow the SB best practices AFAIU in this case. Thanks

examples/camel-example-spring-boot-strimzi/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+
+	<profiles>
+		<profile>
+			<id>kubernetes</id>
+			<build>
+				<plugins>
+					<plugin>
+						<artifactId>fabric8-maven-plugin</artifactId>
+						<groupId>io.fabric8</groupId>
+						<version>${fabric8-maven-plugin-version}</version>
```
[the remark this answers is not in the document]
mabulgu: will do this in another PR as suggested.

examples/camel-example-spring-boot-strimzi/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+			<artifactId>spring-boot-starter-web</artifactId>
+		</dependency>
+
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-actuator</artifactId>
+		</dependency>
+
+		<!-- Camel -->
+		<dependency>
```
davsclaus: This should not be needed as you just depend on the camel-spring-boot-starter

examples/camel-example-spring-boot-strimzi/src/main/java/org/apache/camel/example/strimzi/Application.java
```diff
@@ -0,0 +1,32 @@
+/**
+ *  Copyright 2005-2018 Red Hat, Inc.
```
davsclaus: Use ASF license header, not Red Hat

examples/camel-example-spring-boot-strimzi/src/main/java/org/apache/camel/example/strimzi/StrimziRouteBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.camel.component.kafka.KafkaConstants;
+import org.springframework.stereotype.Component;
+
+@Component
+public class StrimziRouteBuilder extends RouteBuilder {
+
+    @Override
+    public void configure() throws Exception {
+
+        from("timer://foo?period=1000")
```
davsclaus: Indent the route DSL so it looks nicer

examples/camel-example-spring-boot-strimzi/src/main/java/org/apache/camel/example/strimzi/StrimziRouteBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+        .routeId("Producer Route")
+        .setBody(simple("Hi, this is Camel-Strimzi example from {{environment}} environment"))
+        .to("{{camel.component.kafka}}");
```
davsclaus: This is wrong, you would still say "kafka:TestLog"

examples/camel-example-spring-boot-strimzi/src/main/resources/application-local.properties
```diff
@@ -0,0 +1,2 @@
+environment=local
+camel.component.kafka=kafka:TestLog?brokers=KAFKA_BOOTSTRAP_ROUTE_URL&sslEndpointAlgorithm=HTTPS&securityProtocol=SSL&sslTruststoreLocation=truststore.jks&sslTruststorePassword=123456
```
davsclaus: Here you configure multiple options one by one

camel.component.kafka.brokers=XXX
camel.component.kafka.security-protocol=SSL

See this doc page for all the option names
https://github.com/apache/camel-spring-boot/blob/master/components-starter/camel-kafka-starter/src/main/docs/kafka-starter.adoc

mabulgu: Ah ok. it is about the naming then. I misunderstood. That's why I was skeptical about this.

PR #70, opened 2020-05-14
Widget gadget example

akoufoudakis:
Classical widget-gadget integration example from the EIP book, running on SpringBoot and using AMQP and JSONPath.


oscerd:
Thanks

PR #101, opened 2020-07-03
fix (Environment): multiple environments

UniqueDong:
Fix multiple environments when used with dubbo framework causing boot failure. 
the dubbo class org.apache.dubbo.spring.boot.autoconfigure.DubboRelaxedBinding2AutoConfiguration#dubboScanBasePackagesPropertyResolver create a PropertyResolver causing boot failure.

### DIFF
--- a/core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/SpringPropertiesParser.java
+++ b/core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/SpringPropertiesParser.java
@@ -18,15 +18,14 @@ package org.apache.camel.spring.boot;
 
 import org.apache.camel.component.properties.DefaultPropertiesParser;
 import org.apache.camel.component.properties.PropertiesLookup;
-import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.EnvironmentAware;
+import org.springframework.core.env.Environment;
 import org.springframework.core.env.PropertyResolver;
 
-class SpringPropertiesParser extends DefaultPropertiesParser {
+class SpringPropertiesParser extends DefaultPropertiesParser implements EnvironmentAware {
 
     // Members
 
-    @Autowired
-    @Qualifier("environment")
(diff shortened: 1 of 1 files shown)


UniqueDong:
> It's not completely clear what you are doing. Dubbo is not among the camel-spring-boot starters, we don't have support for it. Please explain a bit better the scenario and why camel-spring-boot should be involved.

My program running environment is dubbo. Now I need to introduce the camel-ftp component to read the ftp protocol data and execute the rpc call through dubbo. Since dubbo also created a PropertyResolver to prevent the program from starting, it is specified by @Qualifier("environment") The propertyResolver property of org.apache.camel.spring.boot.SpringPropertiesParser uses Spring's StandardEnvironment

oscerd:
I don't think this is the correct way or at least I don't think this is a scenario we should consider here. @davsclaus what do you think?

davsclaus:
Yeah I would like to know some more about what affect this PR does to everyone else.

UniqueDong:
> I don't think this is the correct way or at least I don't think this is a scenario we should consider here. @davsclaus what do you think?
How about the following solutions？Implement EnvironmentAware to obtain the Spring Environment, so that the PropertyResolver injected by Spring can be implemented without the PropertyResolver created by other components
```java
class SpringPropertiesParser extends DefaultPropertiesParser implements EnvironmentAware {

    // Members

    private PropertyResolver propertyResolver;
	
	@Override
    public void setEnvironment(Environment environment) {
        this.propertyResolver = environment;
    }

    // Overridden

    @Override
    public String parseProperty(String key, String value, Properties properties) {
        return propertyResolver.getProperty(key);
    }

}
```

oscerd:
I still don't think this is the right place for this. We never claim we are supporting dunno, so this is something useless for camel-spring-boot user base. This is an hack you should do in your application. Just my pov

oscerd:
Since there wasn't more feedback on this and since there wasn't a clarification, I'm closing this one.

PR #103, opened 2020-07-04
[CAMEL-14565] Add additional metadata json file to include description docs for spring properties

amdmdi:
Created additional-spring-configuration-metadata.json to add description
to properties inherited from parent configuration classes.
This is because spring-boot-configuration-processor detects inherited
properties by their public getters/setters but can not find the corresponding
javadoc used for generated adoc in the parent classes.

This is the second possible solution for issue CAMEL-14565. Another is copy fields from
parent classes to include the javadoc (and default values)

I created a second pull request for **camel-zookeeper-starter** using the other solution.

Please advise on which solution is better.


amdmdi:
See also PR #104

amdmdi:
PS: I first created this file by using jq on the one spring processor generates :

cat target/classes/META-INF/spring-configuration-metadata.json | jq ' .properties | map(select(has("description") | not)) | .[] += {description: "TODO", defaultValue: "TODO"} | {properties: .}'

davsclaus:
Lets go with the other javadoc style.

amdmdi:
OK, I created PR #105 with the javadoc solution for **camel-consul-starter** too.

PR #104, opened 2020-07-04
[CAMEL-14565]: Copied fields from parent class to include javadoc description of spring config properties

amdmdi:
Copied properties from parent classes to include javadoc because
spring-boot-configuration-processor detects inherited properties by their public
getters/setters but can not find the corresponding javadoc used for generated adoc
in the parent class.

This is the first possible solution for issue CAMEL-14565. Another is to add  a metadata
file **META-INF/additional-spring-configuration-metadata.json** with the proper description.

I created a second pull request for **camel-consul-starter** using the second solution.

Please advise on which solution is better.


amdmdi:
See also PR #103

amdmdi:
PS: I first created this file is by using jq on the one spring processor generates :  

`cat target/classes/META-INF/spring-configuration-metadata.json | jq ' .properties | map(select(has("description") | not)) | .[] += {description: "TODO", defaultValue: "TODO"} | {properties: .}'`

davsclaus:
I like this the best as this is easier to maintain as we can just copy if for example a new option is added in zookeeper etc.

PR #114, opened 2020-07-19
"CAMEL-15274:camel-spring-boot - Add back http endpoint for route inf…

ramu11:
Another PR already raised  "https://github.com/apache/camel/pull/4017"  to get the route dump for the http endpoints

### DIFF
--- a/core/camel-spring-boot/pom.xml
+++ b/core/camel-spring-boot/pom.xml
@@ -129,7 +129,7 @@
         <dependency>
             <groupId>org.apache.camel</groupId>
             <artifactId>camel-management</artifactId>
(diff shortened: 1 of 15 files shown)


davsclaus:
The route dump should be optional and only when JMX is enabled, so do not change the pom file. As this allows spring boot users to not use JMX (JMX is disabled by spring boot also).

Also at first you can just remove the route dump as its not the essential information, its the route status.

ramu11:
Hi Davsclaus,

ok ,I will remove the route dump functionality.

ramu11:
I am closing this PR to raise a new one

PR #123, opened 2020-07-27
CAMEL-15334 Camel-CassandraQL Spring Boot integration test is failing…

JiriOndrusek:
… after upgrade to Cassandra driver 4.x

Issue: https://issues.apache.org/jira/browse/CAMEL-15334

Version of datastax driver came from spring-boot BOM, with value 4.6.1
Camel is using 4.7.2. It has to be overwritten.


JiriOndrusek:
@oscerd I've missed this ignored annotation, which is the reason why I didn't see any failure. I'll finish fix and un-ignore test.

JiriOndrusek:
@oscerd now it should be successful and test is not ignored

PR #192, opened 2020-10-08
Removal of duplicate keywords

NiteshKoushik:
Removal of duplicate keywords


davsclaus:
You need to fix this in the source code, do a git grep and you can find where it has the same text and where to fix in the source code

oscerd (changes requested):
Needs to be done in the source code

NiteshKoushik:
Thanks for the suggestion.
I cannot clone and download the code in my local as there are some security restrictions - when it comes to the organization limits. I can download the zip code, but that will not help in finding out the exact location where I need to change.

Is it possible to share the exact page, where I need to change, so that I can make the respective edits and update it.

davsclaus:
This is actually inheirted from camel core in core/camel-main/src/main/java/org/apache/camel/main/DefaultConfigurationProperties.java
https://github.com/apache/camel/blob/master/core/camel-main/src/main/java/org/apache/camel/main/DefaultConfigurationProperties.java

davsclaus:
At this line
https://github.com/apache/camel/blob/master/core/camel-main/src/main/java/org/apache/camel/main/DefaultConfigurationProperties.java#L1486

NiteshKoushik:
Thanks @davsclaus 👍 
Closing the request as the same changes done in PR https://github.com/apache/camel/pull/4416

PR #222, opened 2020-11-03
Added Spring Boot XML DSL archetype

fg78nc:


oscerd (approved):

fg78nc:
HI,

According to documentation Camel will pick up XML context/route definition
from the classpath.
However it did not work, unless I explicitly add @ImportResource.
It did not even work after adding camel.springboot.xml-routes to
application.properties.
Is it expected behavior?

oscerd:
This is a good question, we did some change after 3.4.x LTS. @davsclaus can you shed some light on XML routes in SB?

davsclaus:
Its 2 different thing

spring xml files with <beans> are ONLY spring that can load this via that spring boot @import resource annotation.

The xml routes is ONLY for Camel XML files which are <routes> xml snippets (not spring at all).

fg78nc:
Oh, I see now.
I will try that. Where can I see an example of XML structure of Camel xml
route?
Is it just a route definition, similar to what can be dropped into ActiveMQ?
Also probably documentation has to be updated in several places, including
https://camel.apache.org/manual/latest/xml-configuration.html and
https://camel.apache.org/components/latest/spring-summary.html ?

davsclaus:
The old world with spring xml files (eg beans) is not something for spring boot. So we are not looking for a maven archetype that does this.

fg78nc:
I agree with you, it is not an idiomatic way for modern Spring Boot
development.
However please consider use cases where a user wants to migrate their Camel
application running on Fuse/Karaf to Spring Boot Java DSL.

fg78nc:
@davsclaus @oscerd Please review and make a decision.

oscerd:
I think we should close this without merging.

oscerd:
Thanks anyway.

PR #260, opened 2021-01-19
[CAMEL-16063]should consider multiple ApplicationContext instances wh…

ffang:
…en specifying another management.server.port


oscerd (approved):

ffang:
Thanks @oscerd !

PR #276, opened 2021-02-03
CAMEL-15560: generic route loader

lburgazzoli:


oscerd (approved):

michaelpearce-gain:
This seems to have broken spring boot loading of camel xml's once packaged. route xml's load fine in ide, but once packaged into a springboot runtime jar/war they are no longer loaded. 3.7.6 works fine, 3.8.0 (this change) it stops working.

PR #407, opened 2021-11-10
Camel spring boot 3.12.x collect

djencks:
See apache/camel-website#667

Replaces copying with aggregate collector.
Generates options in spring-boot.adoc.


davsclaus:
3.12.x will be removed when we release 3.13.x

PR #415, opened 2021-11-25
CAMEL-17274 - Fix regression caused by CAMEL-15560

michaelandrepearce:
This fixes regression caused by CAMEL-15560 removal of collectXmlRoutesFromDirectory methods, which overrode default to use spring's application context, but no equivalent was made in replacement.

The fix basically implements new equivalent method, re-introducing loading resource by the applicationContext

Unfortunately it seems this is only seen/issue up once a fully packaged springboot jar/war is created and run, and route xml's fail to load, as such very hard to unit test, and like wise no pre-existing tests clearly for old behaviour either, thus it broke.

related to: https://stackoverflow.com/questions/68065987/apache-camel-routes-in-xml-not-loaded-in-spring-boot/

### DIFF
--- a/core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/SpringBootRoutesCollector.java
+++ b/core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/SpringBootRoutesCollector.java
@@ -16,20 +16,27 @@
  */
 package org.apache.camel.spring.boot;
 
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.Set;
 
 import org.apache.camel.CamelContext;
 import org.apache.camel.ExtendedCamelContext;
 import org.apache.camel.RoutesBuilder;
+import org.apache.camel.RuntimeCamelException;
 import org.apache.camel.builder.LambdaRouteBuilder;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.main.DefaultRoutesCollector;
 import org.apache.camel.util.AntPathMatcher;
 import org.apache.camel.util.ObjectHelper;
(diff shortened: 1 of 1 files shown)


mpoutlook:
@davsclaus anything in addition needed? This is blocking upgrading to later versions of camel. Due to the break caused in 3.8

michaelandrepearce:
@davsclaus anything in addition needed? This is blocking upgrading to later versions of camel. Due to the break caused in 3.8

davsclaus:
How are you packing your spring-boot apps - if its a fat-jar then the `FatJarPackageScanResourceResolver` should be the one that finds the files. Can you look into this. As your PR will due to code override not use that resolver.

davsclaus:
In other words your change may cause other problems, although it may be that spring application context can load resources inside fat-jars, which you may assume it can.

And if you have a sample project that reproduces this - that would help

davsclaus:
And create a new JIRA ticket about this, and in that you can link/refer to CAMEL-15560. And update your commit message to use the new JIRA number.

michaelandrepearce:
No the fat jar resolver isnt picking up correctly, you have to use the spring context (as the code use to) and thus why once packaged with spring 3.8 is broken

michaelandrepearce:
Will zip up a sample spring boot app to demonstrate  and attach to jira

davsclaus:
We plan to cut 3.14 by end of next week, so it would be good to have this start of next week so time to make it into the release.

michaelandrepearce:
Ill get it done Monday when back in work

michaelandrepearce:
@davsclaus i had some free time this evening, as such created JIRA:
https://issues.apache.org/jira/browse/CAMEL-17274

Extracted sample application demonstrating the issue (not the app is set to 3.7.6 showing its working, but breaks when you upgrade to 3.8.0 or above), this is zipped and attached to JIRA ticket.

Amended commit message to new JIRA id.

davsclaus:
Found another way

michaelpearce-gain:
@davsclaus thanks.

PR #481, opened 2022-03-25
CAMEL-17846 add tests in camel-aws2-ddb-starter

JiriOndrusek:
Issue: https://issues.apache.org/jira/browse/CAMEL-17846

Added tests for all supported operation of aws2-ddb-producer (see [documentation](https://camel.apache.org/components/3.15.x/aws2-ddb-component.html#_supported_producer_operationshttps://camel.apache.org/components/3.15.x/aws2-ddb-component.html#_supported_producer_operations))

Added basic test to cover ddbastreams.

### DIFF
--- a/components-starter/camel-aws2-ddb-starter/pom.xml
+++ b/components-starter/camel-aws2-ddb-starter/pom.xml
@@ -47,6 +47,18 @@
       </exclusions>
       <!--END OF GENERATED CODE-->
     </dependency>
+    <dependency>
+      <groupId>org.apache.camel</groupId>
+      <artifactId>camel-test-infra-aws-v2</artifactId>
+      <version>${camel-version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <scope>test</scope>
+    </dependency>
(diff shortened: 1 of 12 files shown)


components-starter/camel-aws2-ddb-starter/pom.xml
```diff
@@ -47,6 +47,18 @@
       </exclusions>
       <!--END OF GENERATED CODE-->
     </dependency>
+    <dependency>
+      <groupId>org.apache.camel</groupId>
+      <artifactId>camel-test-infra-aws-v2</artifactId>
+      <version>${project.version}</version>
```
davsclaus, line 53: @JiriOndrusek when you add JARs that come from core camel project, then you MUST use ${camel.version} as the version placeholder - otherwise we cannot release - we had trouble over the weekend when the release failed.

davsclaus (changes requested):

components-starter/camel-aws2-ddb-starter/pom.xml
JiriOndrusek: Thanks for spotting it! I'll fix that

JiriOndrusek: @davsclaus it is fixed ow

davsclaus (approved):

PR #529, opened 2022-04-16
Use commons-lang for aws2-sqs-starter

cunningt:
Was getting a class not found error on org.testcontainers.shaded.org.apache.commons.lang.RandomStringUtils - replace with commons-lang version and add in commons-lang dependency

### DIFF
--- a/components-starter/camel-aws2-sqs-starter/pom.xml
+++ b/components-starter/camel-aws2-sqs-starter/pom.xml
@@ -48,6 +48,11 @@
       <!--END OF GENERATED CODE-->
     </dependency>
     <dependency>
+      <groupId>commons-lang</groupId>
+      <artifactId>commons-lang</artifactId>
+      <version>${commons-lang-version}</version>
+    </dependency>
+    <dependency>
       <groupId>org.apache.camel</groupId>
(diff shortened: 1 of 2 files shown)


components-starter/camel-aws2-sqs-starter/pom.xml
```diff
@@ -47,6 +47,11 @@
       </exclusions>
       <!--END OF GENERATED CODE-->
     </dependency>
+    <dependency>
+      <groupId>commons-lang</groupId>
+      <artifactId>commons-lang</artifactId>
+      <version>${commons-lang-version}</version>
```
davsclaus, line 53: It seems its only used for testing, can you use <scope>test</scope>
And can you maybe use commons lang3 as the one you use now is legacy

davsclaus (changes requested):

components-starter/camel-aws2-sqs-starter/pom.xml
cunningt: Changed scope to test, changed to use commons-lang3 version

davsclaus (approved):

PR #538, opened 2022-04-22
Archive surefire logs to try to diagnose integration test failures

cunningt:
We're getting failures in some of the integration tests but the build.log stored by the PR job does not have a lot of detail to go on -

Example, from CamelSqlTest failure in one of the recent PR build runs -
```
[ERROR] org.apache.camel.itest.springboot.CamelSqlTest.componentTests  Time elapsed: 25.465 s  <<< ERROR!
java.lang.Exception: 
Caused by: org.opentest4j.AssertionFailedError: Some unit tests failed (1/25), check the logs for more details
```

I'd like to temporarily start archiving the surefire logs to see if we could diagnose the failures - I'm not sure whether this is going to give us additional information or not.    I haven't been able to reproduce these failures locally.     

I think we can expire the logs pretty quickly so that they don't take up a lot of space.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -46,3 +46,12 @@ jobs:
       with:
         name: build.log
         path: build.log
+    - name: Archive Surefire Logs
+      uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: integration-test-surefire-logs
+        path: |
+          tests/camel-itest-spring-boot/target/surefire-reports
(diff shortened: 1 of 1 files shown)


davsclaus (approved):

davsclaus:
@otavioprado any comments as you also worked on some log archive from gh actions in the core project

orpiske:
Hi, +1 from me. Indeed, we are doing the same on core.

PR #540, opened 2022-04-24
Filter out annotated spring boot tests

Croway:
This PR should fix failures in gh actions, in particular newest tests added under components-starter shouldn't be executed as integration tests are already executed during components build via @CamelSpringBootTest and @SpringBootTest annotations.

That's why I've added a new filter that won't execute tests classes annotated with springboot test annotations.

Anyway, the failure was caused due to flaky tests in zipfile and sql starter components. I will investigate those.

@cunningt


oscerd (approved):
Needs rebase but looks good

davsclaus (approved):

PR #556, opened 2022-05-06
Set pom type to parent deps

Croway:
Set correct type to camel-*-parent flattened dependencies.

@cunningt 

@davsclaus @oscerd Hello, I remember some discussions around camel-spring-boot-dependencies bom deprecation, do you still think that camel-spring-boot-bom should be used instead of -dependencies?


davsclaus (approved):

davsclaus:
Yes SB users want to use SB as first class, and then Camel -starter as addons.
You can say the same for CEQ, where Quarkus users want to use Q first class and then Camel as extensions

PR #577, opened 2022-06-25
(chores) ci: upgrade github actions to the same versions used by Camel Core

orpiske:


orpiske:
The failure is unrelated to this, so it's safe to merge.

PR #580, opened 2022-06-25
(chores) ci: included dependency review

orpiske:
See apache/camel#7826


orpiske:
The failure is unrelated to this, so it's safe to merge.

PR #582, opened 2022-06-27
Update to SB 2.7.1

Croway:
https://spring.io/blog/2022/06/23/spring-boot-2-7-1-available-now was released

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,7 +108,7 @@
         <compiler.fork>false</compiler.fork>
 
         <!-- Spring-Boot target version -->
-        <spring-boot-version>2.7.0</spring-boot-version>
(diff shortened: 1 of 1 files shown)


Croway:
I've executed mllp tests locally and they are succesful, is it possible to retrigger the PR build? @orpiske

davsclaus:
There is another place where 2.7.0 is defined, you can find it via `git grep 2.7.0`

Croway:
> There is another place where 2.7.0 is defined, you can find it via `git grep 2.7.0`

I couldn't find another place with 2.7.0, but if I remember correctly sometime ago we changed this behaviour with https://github.com/apache/camel-spring-boot/commit/5d5d45a734515b012b7e6f9a52cfc7c004603228 

Am I missing something? @davsclaus

PR #621, opened 2022-09-02
✨ CAMEL-18449 platform-http servlet implementation

Croway:
The idea behind this PR is to use camel-platform-http-starter and camel-servlet-starter without any additional configuration.
If camel-servlet is configured in the context (camel-servlet-starter is declared as a dependency), then a PlatformHttpEngine implementation is created using the servlet, I'm not 100% sure about the reflection, if you think it can be done in some other way please let me know.

### DIFF
--- a/components-starter/camel-platform-http-starter/pom.xml
+++ b/components-starter/camel-platform-http-starter/pom.xml
@@ -40,6 +40,22 @@
       <version>${camel-version}</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.camel</groupId>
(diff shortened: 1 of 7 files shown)


davsclaus:
You can have dependency on camel-servlet that is okay, then you do not need to use reflection

Croway:
Thank you @davsclaus, I've updated it and created the following example https://github.com/apache/camel-spring-boot-examples/pull/83

davsclaus (approved):

PR #654, opened 2022-10-26
Remove authSource param from MongoDB tests

mcarlett:


oscerd (approved):

davsclaus:
Can you tell a bit why this should be removed?

mcarlett:
this parameter was failing in the older version but in the main branch it works, sorry about this PR, it is not necessary anymore

PR #682, opened 2022-12-08
Fixed Flaky Test Caused by JSON permutations

bmk15897:
### Description
Flaky Tests found using [NonDex](https://github.com/TestingResearchIllinois/NonDex) by running the commands -
`
mvn edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=org.apache.camel.component.gson.springboot.GsonMarshalExclusionTest#testMarshalAndUnmarshalPojoWithExclusion
`
The logged failure was-
```
[WARNING] Flakes:
[WARNING] org.apache.camel.component.gson.springboot.GsonMarshalExclusionTest.testMarshalAndUnmarshalPojoWithExclusion
[ERROR]   Run 1: GsonMarshalExclusionTest.testMarshalAndUnmarshalPojoWithExclusion:80 expected: <{"age":30,"height":190}> but was: <{"height":190,"age":30}>
```
### Investigation
The tests  fail with a comparison error while comparing an expected JSON String and the result from the value returned by the funtion template.requestBody(...). The function of the that converts given Pojo Object to String does not guarantee the order of the fields in the Pojo. This makes the test outcome non-deterministic, and the test fails whenever the function returns a mismatch in order of the elements in the JSON String. To fix this, the expected and actual keys should be checked in a more deterministic way so that the assertions do not fail.

### Fix
Expected and actual arrays can be compared without making assumptions about the order of the keys in these JSON Strings. Instead of using assertEquals for the complete JSON String, using contains for every subpart of the expected String makes the test more deterministic and ensures that the flakiness from the test is removed.

The PR does not introduce a breaking change.

### DIFF
--- a/components-starter/camel-gson-starter/src/test/java/org/apache/camel/component/gson/springboot/GsonMarshalExclusionTest.java
+++ b/components-starter/camel-gson-starter/src/test/java/org/apache/camel/component/gson/springboot/GsonMarshalExclusionTest.java
@@ -79,7 +79,9 @@ public class GsonMarshalExclusionTest {
 
         Object marshalled = template.requestBody("direct:inPojoExcludeWeight", in);
         String marshalledAsString = context.getTypeConverter().convertTo(String.class, marshalled);
-        assertEquals("{\"age\":30,\"height\":190}", marshalledAsString);
+        assertTrue(marshalledAsString.contains("\"height\":190"));
+        assertTrue(marshalledAsString.contains("\"age\":30"));
+		assertFalse(marshalledAsString.contains("\"weight\":70"));
 
         template.sendBody("direct:backPojoExcludeWeight", marshalled);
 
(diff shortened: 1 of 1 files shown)


bmk15897:
@davsclaus Can You please review this?

components-starter/camel-gson-starter/src/test/java/org/apache/camel/component/gson/springboot/GsonMarshalExclusionTest.java
```diff
@@ -79,7 +79,9 @@ public void testMarshalAndUnmarshalPojoWithExclusion() throws Exception {
 
         Object marshalled = template.requestBody("direct:inPojoExcludeWeight", in);
         String marshalledAsString = context.getTypeConverter().convertTo(String.class, marshalled);
-        assertEquals("{\"age\":30,\"height\":190}", marshalledAsString);
+        assertTrue(marshalledAsString.contains("\"height\":190"));
+        assertTrue(marshalledAsString.contains("\"age\":30"));
+		assertFalse(marshalledAsString.contains("\"weight\":70"));
```
davsclaus, line 84: Fix indent so the code is aligned like the code above

bmk15897: Fixed. Apologies for the inconvenience.

davsclaus (approved):

PR #704, opened 2023-01-17
CAMEL-18932: added missing log4j versions

orpiske:


orpiske:
@davsclaus do you mind a second look, please? This should fix the build break, but I don't want to interfere w/ other activities for CAMEL-18932.

oscerd (approved):

orpiske:
Obs.: failing because the `camel-dependencies` pom in the cache is using version 2.18.0 for log4j, which seems unavailable.

davsclaus:
you should use 2.19.0 version of log4j

davsclaus:
I assume this is working now and we can close this PR ?

orpiske:
> you should use 2.19.0 version of log4j

i am not suing any version here ... it's taking from Camel. Do you mean I should bypass it?

orpiske:
> I assume this is working now and we can close this PR ?

I haven't tried it, but if it's working now, let's close it.

PR #711, opened 2023-01-19
Update to Spring Boot 2.7.8

gzurowski:


davsclaus (approved):

davsclaus:
@gzurowski if you have time then check these dependency upgrades from SB 2.7.8 and find them in the camel core project and try to upgrade the same or higher version

https://github.com/spring-projects/spring-boot/releases/tag/v2.7.8

such as micrometer, netty, etc.

oscerd (approved):

PR #793, opened 2023-03-21
CAMEL-19177: camel-platform-http - Spring Boot implementation that us…

davsclaus:
…e directly the HTTP server


davsclaus:
This requires a new SNAPSHOT build of core as there are some commits this requires

PR #805, opened 2023-03-31
CAMEL-15105: adjust CamelContext configuration

orpiske:


oscerd (approved):

orpiske:
Build is going to fail for a while due to relying on unpublished code, but it should be fine, so merging ...

PR #837, opened 2023-05-11
CAMEL-18963: upgrade both spring-boot starters using camel-openapi-java

klease:
Modify the AutoConfiguration classes to use the modified return type in RestOpenApiReader.read method. Since OpenAPI is now directly returned there is no need to convert it to json and reparse it with swagger. In addition, this change allows handling of an OpenAPI 2.x specification which will be automatically upgraded to OpenAPI 3.x. Previously this would have failed due to the direct use of OpenAPIV3Parser.


klease:
Waiting for the core camel openapi components to be built.

klease:
The PR build fails in the camel-kamelet-starter due to a failing test. See CAMEL-19345.

klease:
Merci Claus.

PR #844, opened 2023-05-26
CAMEL-13573 : create parquet-avro dataformat 

onderson:


oscerd (changes requested):
This is not complete, I'll have a look.

PR #868, opened 2023-06-20
CAMEL-19475: re-add ActiveMQ starter

orpiske:


oscerd (approved):

davsclaus (approved):

orpiske:
> The build failure is suspicious ... I am investigating it.

Oh, it's failing because the `camel-activemq` is missing on the snapshot repos. We have to wait for publication.

PR #936, opened 2023-09-12
CAMEL-19851: camel-http - Allow to configure timeouts natively

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-19851 for 4.1

## Motivation

The timeout can only be configured like any other complex type, indeed, we need to define a dedicated bean in our Spring context and refer to that bean in the configuration of our component which could be simplified by supporting natively timeouts.

## Modifications:

* Add new converters to support natively a timeout in milliseconds expressed as long, integer, or string

### DIFF
--- a/components-starter/camel-http-starter/src/main/java/org/apache/camel/component/http/springboot/HttpComponentTimeoutConverter.java
+++ b/components-starter/camel-http-starter/src/main/java/org/apache/camel/component/http/springboot/HttpComponentTimeoutConverter.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.http.springboot;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import org.apache.hc.core5.util.Timeout;
+import org.springframework.boot.context.properties.ConfigurationPropertiesBinding;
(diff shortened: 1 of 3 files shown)


components-starter/camel-http-starter/src/main/java/org/apache/camel/component/http/springboot/HttpComponentTimeoutConverter.java
```diff
@@ -0,0 +1,60 @@
...
+@ConfigurationPropertiesBinding
+@Component
+public class HttpComponentTimeoutConverter extends HttpComponentConverter {
+
+    @Override
+    public Set<ConvertiblePair> getConvertibleTypes() {
+        Set<ConvertiblePair> answer = new LinkedHashSet<>();
+        answer.add(new ConvertiblePair(Integer.class, org.apache.hc.core5.util.Timeout.class));
+        answer.add(new ConvertiblePair(Long.class, org.apache.hc.core5.util.Timeout.class));
+        answer.add(new ConvertiblePair(String.class, org.apache.hc.core5.util.Timeout.class));
```
Akvel, line 39: I think you should also remove converter String->Timeout from [HttpComponentConverter](https://github.com/apache/camel-spring-boot/blob/main/components-starter/camel-http-starter/src/main/java/org/apache/camel/component/http/springboot/HttpComponentConverter.java#L51) or we can get  race

components-starter/camel-http-starter/src/main/java/org/apache/camel/component/http/springboot/HttpComponentTimeoutConverter.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.boot.context.properties.ConfigurationPropertiesBinding;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.convert.TypeDescriptor;
+import org.springframework.stereotype.Component;
+
+
+@Configuration(proxyBeanMethods = false)
+@ConfigurationPropertiesBinding
+@Component
+public class HttpComponentTimeoutConverter extends HttpComponentConverter {
```
Akvel, line 32: Please, update doc also - https://github.com/apache/camel-spring-boot/blob/main/components-starter/camel-http-starter/src/main/docs/http.json

essobedo:
@Akvel Those files are generated which is why they cannot be modified. If we have other use cases like this one, I believe that we could consider improving the generator but for now, I'm not sure, it is worth the effort.
@oscerd WDYT?

Akvel:
> I think it's something optional to do. We could open an issue to avoid forgetting but it's not critical imo

I agree that It is minor because most people will be use int/long. Just ligthup this case

essobedo:
Depends on https://github.com/apache/camel-spring-boot/pull/937

PR #961, opened 2023-09-29
CAMEL-19864: renable test after fixing root issue.

klease:


oscerd (approved):

klease:
Build failure due to openapi component which should be OK when Jenkins finishes current Camel core build.

PR #1042, opened 2023-12-13
CAMEL-20231: make generators configurable

Croway:


davsclaus (approved):

oscerd (approved):

davsclaus:
okay had to do a few small changes - this -starter is somewhat special. It does some weird custom binder stuff, that should not be needed, but do it more normal spring boot way.

Croway:
> okay had to do a few small changes - this -starter is somewhat special. It does some weird custom binder stuff, that should not be needed, but do it more normal spring boot way.

Interesting, I am sorry I thought it was an easy change, but doesn't `private String randomIvGeneratorAlgorithm;` resolve to null if not set?

davsclaus:
yes but isnt there a default algorithm in jasypt itself. And if it has a default then it can also be set in that field, then SB and tooling know what the default value is.

PR #1097, opened 2024-02-29
feat: add initial support for camel-k

lburgazzoli:
Requires https://github.com/apache/camel/pull/13376

### DIFF
--- a/core/camel-spring-boot-xml/src/main/java/org/apache/camel/spring/boot/xml/SpringBootXmlCamelContextConfigurer.java
+++ b/core/camel-spring-boot-xml/src/main/java/org/apache/camel/spring/boot/xml/SpringBootXmlCamelContextConfigurer.java
@@ -22,12 +22,10 @@ import org.apache.camel.spi.Injector;
 import org.apache.camel.spring.SpringCamelContext;
 import org.apache.camel.spring.boot.CamelAutoConfiguration;
 import org.apache.camel.spring.boot.CamelConfigurationProperties;
-import org.apache.camel.spring.spi.SpringInjector;
 import org.apache.camel.spring.xml.XmlCamelContextConfigurer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.ApplicationContext;
-import org.springframework.context.ConfigurableApplicationContext;
 
 /**
  * Used to merge Camel Spring Boot configuration with {@link org.apache.camel.CamelContext} that
--- a/core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/CamelAutoConfiguration.java
+++ b/core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/CamelAutoConfiguration.java
@@ -31,7 +31,6 @@ import org.apache.camel.RuntimeCamelException;
 import org.apache.camel.component.properties.PropertiesComponent;
 import org.apache.camel.component.properties.PropertiesParser;
 import org.apache.camel.main.DefaultConfigurationConfigurer;
-import org.apache.camel.main.DefaultRoutesCollector;
 import org.apache.camel.main.RoutesCollector;
 import org.apache.camel.model.Model;
 import org.apache.camel.spi.BeanRepository;
--- a/core-starter/camel-k-starter/pom.xml
+++ b/core-starter/camel-k-starter/pom.xml
@@ -1,0 +1,131 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one or more
+    contributor license agreements.  See the NOTICE file distributed with
+    this work for additional information regarding copyright ownership.
+    The ASF licenses this file to You under the Apache License, Version 2.0
+    (the "License"); you may not use this file except in compliance with
+    the License.  You may obtain a copy of the License at
+
+         http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.camel.springboot</groupId>
+    <artifactId>core-starter</artifactId>
+    <version>4.5.0-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>camel-k-starter</artifactId>
+  <packaging>jar</packaging>
+  <name>Camel SB Starters :: camel-k</name>
+
+  <dependencies>
+
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.camel.springboot</groupId>
+      <artifactId>camel-core-starter</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.camel.springboot</groupId>
+      <artifactId>camel-kubernetes-starter</artifactId>
+    </dependency>
+
+    <!-- Testing dependencies -->
+    <dependency>
+      <groupId>org.apache.camel</groupId>
+      <artifactId>camel-test-spring-junit5</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.camel</groupId>
+          <artifactId>camel-spring-xml</artifactId>
+        </exclusion>
+      </exclusions>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <scope>test</scope>
+    </dependency>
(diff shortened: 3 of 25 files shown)


core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/CamelAutoConfiguration.java
```diff
@@ -189,6 +197,12 @@ public static CamelContext doConfigureCamelContext(ApplicationContext applicatio
         // and call after all properties are set
         DefaultConfigurationConfigurer.afterPropertiesSet(camelContext);
 
+        // apply custom configurations if any
+        // TODO: this must be evaluated as I don't know if it makes sense here.
+        contextCustomizers.stream()
+                .sorted(Comparator.comparing(CamelContextCustomizer::getOrder))
+                .forEach(c -> c.configure(camelContext));
```
lburgazzoli, line 204: @davsclaus I added this to avoid polluting this `camelContext` method, I'm not sure if there is a better way of doing it.

davsclaus: Yeah its something you added, and its not much in use. In camel-core we use it for yaml-dsl but then its internally handled there, and not something that requires camel to trigger on startup.

I am look at that mapstruct thing now ... curious why its there

lburgazzoli: I think I can do something similar with bean post processors but this context customizer does not look too bad

Let me know if you think I should rather explore bean post processors of it this looks ok-ish

davsclaus: But this makes us wonder, should we make `CamelContextCustomizer` a general thing and let it be part of camel-core (camel-main) during bootstrap so it checks if there are custom in the registy and call them on starutp

davsclaus: I would likely try to change this in camel-core via

        // lookup and configure SPI beans
        DefaultConfigurationConfigurer.afterConfigure(camelContext);

davsclaus: That method does lookup various beans and configure them. I would think `CamelContextCustomizer` should be there instead.

Also if you think to make it cleaner, we can add a new method on `DefaultConfigurationConfigurer` just for these customizers, and invoke it after the method above. And then you also need to remember to call this method from Camel Spring Boot (and the same for Quarkus unless it does some special magic for this)

davsclaus: Otherwise then I think the rest is good

core/camel-spring-boot-xml/src/main/java/org/apache/camel/spring/boot/xml/SpringBootXmlCamelContextConfigurer.java
```diff
@@ -47,7 +49,9 @@ public void configure(ApplicationContext applicationContext, SpringCamelContext
             // spring boot is not capable at this phase to use an injector that is creating beans
             // via spring-boot itself, so use a default injector instead
             camelContext.setInjector(new DefaultInjector(camelContext));
-            CamelAutoConfiguration.doConfigureCamelContext(applicationContext, camelContext, config);
+
+            // TODO: not sure how this works TBH
+            CamelAutoConfiguration.doConfigureCamelContext(applicationContext, camelContext, List.of(), config);
```
davsclaus, line 54: Yeah this is for old legacy Spring XML with embedded `<camelContext>` in Spring `<beans>` XML files. that you run inside spring boot.

core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/CamelAutoConfiguration.java
```diff
@@ -95,19 +99,23 @@ public class CamelAutoConfiguration {
...
+
+        return doConfigureCamelContext(applicationContext, camelContext, contextCustomizers, config);
     }
 
     /**
      * Not to be used by Camel end users
      */
     public static CamelContext doConfigureCamelContext(ApplicationContext applicationContext,
                                                 CamelContext camelContext,
+                                                List<CamelContextCustomizer> contextCustomizers,
```
davsclaus, line 118: This can then be removed if we make camel-core do the customizer itself, as then it will just lookup them from the registry like all the other beans it does today (same same)

core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/CamelAutoConfiguration.java
lburgazzoli: ok, so I'll try to move to camel-main, the only thing I have to check is on camel-quarkus as it is heavily used

lburgazzoli: moved to camel-main: https://github.com/apache/camel/pull/13376

core/camel-spring-boot-xml/src/main/java/org/apache/camel/spring/boot/xml/SpringBootXmlCamelContextConfigurer.java
lburgazzoli: not needed anymore (https://github.com/apache/camel/pull/13376)

core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/CamelAutoConfiguration.java
lburgazzoli: Not needed anymore (moved to camel-main: https://github.com/apache/camel/pull/13376)

lburgazzoli:
Looks like there are some issues related to openapi v2 rmeoval

```
[INFO] Running org.apache.camel.openapi.RestOpenApiReaderTest
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 4.058 s <<< FAILURE! -- in org.apache.camel.openapi.RestOpenApiReaderTest
[ERROR] org.apache.camel.openapi.RestOpenApiReaderTest.testReaderRead -- Time elapsed: 0.010 s <<< ERROR!
java.lang.IllegalArgumentException: OpenAPI 2.x is not supported
	at org.apache.camel.openapi.RestOpenApiReader.checkCompatOpenApi2(RestOpenApiReader.java:182)
	at org.apache.camel.openapi.RestOpenApiReader.read(RestOpenApiReader.java:176)
	at org.apache.camel.openapi.RestOpenApiReaderTest.testReaderRead(RestOpenApiReaderTest.java:105)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
```

davsclaus:
Are you on old code ?

lburgazzoli:
> Are you on old code ?

No, rebased on main, I see that OpenApi v2 is hardcoded in some of the tests https://github.com/apache/camel-spring-boot/blob/main/components-starter/camel-openapi-java-starter/src/test/java/org/apache/camel/openapi/RestOpenApiModelApiSecurityRequirementsTest.java#L94

davsclaus:
Ah sorry its those darn test copy from core, that have little value and get out of date

davsclaus:
fixing this now

davsclaus:
pushed fix

lburgazzoli:
ok so now failures are because there is no yet a -SNAPSHOT ready

core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/CamelAutoConfiguration.java
```diff
@@ -95,11 +95,14 @@ public class CamelAutoConfiguration {
     @ConditionalOnMissingBean(CamelContext.class)
     CamelContext camelContext(ApplicationContext applicationContext,
                               CamelConfigurationProperties config,
-                              CamelBeanPostProcessor beanPostProcessor) throws Exception {
+                              CamelBeanPostProcessor beanPostProcessor,
+                              List<CamelContextCustomizer> contextCustomizers) throws Exception {
```
davsclaus, line 99: This is no longer needed

core/camel-spring-boot-xml/src/main/java/org/apache/camel/spring/boot/xml/SpringBootXmlCamelContextConfigurer.java
```diff
@@ -47,6 +45,8 @@ public void configure(ApplicationContext applicationContext, SpringCamelContext
             // spring boot is not capable at this phase to use an injector that is creating beans
             // via spring-boot itself, so use a default injector instead
             camelContext.setInjector(new DefaultInjector(camelContext));
+
+            // TODO: not sure how this works TBH
```
davsclaus, line 49: I think this file can be reverted

davsclaus:
There was some leftover in the camel-spring-boot core module. If this gets reverted then this PR is only adding a new camel-k module, and then its perfect fine to merged.

core/camel-spring-boot-xml/src/main/java/org/apache/camel/spring/boot/xml/SpringBootXmlCamelContextConfigurer.java
lburgazzoli: done

core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/CamelAutoConfiguration.java
lburgazzoli: done

davsclaus (approved):

oscerd (approved):

PR #1113, opened 2024-03-19
[CAMEL-20524] Harmonize configuration between spring-boot and camel-main

gnodet:
It seems some javadoc is lagging, I suspect some fixes have been made in spring-boot javadoc and not backported to camel-main...


lburgazzoli:
looks good

PR #1140, opened 2024-05-06
Make Jasypt beans conditional so that can be overridden

Croway:
At the moment, the documentation does not work as is, a bean like
```
<bean id="jasypt" class="org.apache.camel.component.jasypt.JasyptPropertiesParser">
    <!-- password is mandatory, you can prefix it with sysenv: or sys: to indicate it should use
         an OS environment or JVM system property value, so you don't have the master password defined here -->
    <property name="password" value="secret"/>
</bean>
```
has to be defined as primary="true" otherwise two spring beans are loaded in the context for the same class.

Marking beans as conditional solves this issue.

### DIFF
--- a/components-starter/camel-jasypt-starter/src/main/java/org/apache/camel/component/jasypt/springboot/JasyptEncryptedPropertiesAutoconfiguration.java
+++ b/components-starter/camel-jasypt-starter/src/main/java/org/apache/camel/component/jasypt/springboot/JasyptEncryptedPropertiesAutoconfiguration.java
@@ -58,6 +58,7 @@ import static org.springframework.core.annotation.AnnotationUtils.findAnnotation
 public class JasyptEncryptedPropertiesAutoconfiguration {
 
     @Bean
+    @ConditionalOnMissingBean(JasyptEncryptedPropertiesConfiguration.class)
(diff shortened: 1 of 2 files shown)


oscerd (approved):

components-starter/camel-jasypt-starter/src/main/java/org/apache/camel/component/jasypt/springboot/JasyptEncryptedPropertiesAutoconfiguration.java
```diff
@@ -107,6 +109,7 @@ public EncryptablePropertySourcesPlaceholderConfigurer propertyConfigurer(String
      * properties inside the camel context.
      */
     @Bean
+    @ConditionalOnMissingBean(PropertiesParser.class)
```
davsclaus, line 112: I think this could clash with `CamelAutoConfiguration.java` that brings its own parser. It may be that we need to make this jasypt extend the SpringPropertiesParser(env) so it support the default behavior also.

So maybe jasypt properties parser should extend SpringPropertiesParser

davsclaus (approved):

davsclaus:
Can you create a JIRA so we can track this and backport to LTS branches

Croway:
> Can you create a JIRA so we can track this and backport to LTS branches

sure, on it

PR #1199, opened 2024-08-20
CAMEL-19182: camel-platform-http-starter - Async processing with Spring Boot

kulagaIA:
- Spring MVC automatically detects CompletableFuture return type of the handler method and executes asynchronously; `@ResponseBody` on the handler method to disable attempts to resolve mvc view from a response.
- testing async execution with MockMvc
- `@EnableAutoConfiguration` instead of `@SpringBootApplication` on test classes to not scan another tests' configurations

Issue: https://issues.apache.org/jira/browse/CAMEL-19182


components-starter/camel-platform-http-starter/src/test/java/org/apache/camel/component/platform/http/springboot/PlatformHttpBase.java
```diff
@@ -33,7 +32,8 @@ public void testGet() {
     }
 
     @Test
-    public void testPost() {
+    public void testPost() throws InterruptedException {
+        Thread.sleep(500); //wait until http server is up
```
Croway, line 36: could you use Awaitility instead of the sleep?

components-starter/camel-platform-http-starter/src/test/resources/application.properties
```diff
@@ -0,0 +1,2 @@
+logging.level.org.springframework.web: DEBUG
```
davsclaus, line 1: Maybe these should be disabled via `#` so they can quickly be enabled for manual trouble shooting etc

davsclaus (approved):

components-starter/camel-platform-http-starter/src/test/java/org/apache/camel/component/platform/http/springboot/PlatformHttpBase.java
kulagaIA: thanks for your review! Added Awaitility instead of sleep()

Croway (approved):

kulagaIA:
Tests have failed because we run `@DirtiesContext` tests in parallel:
`echo "MVND_OPTS=-V -e -ntp -Dmvnd.threads=2 -Daether.connector.http.connectionMaxTtl=120 -Daether.connector.requestTimeout=300000 -Daether.dependencyCollector.impl=bf -Dmaven.artifact.threads=25 -Dci.env.name=github.com -Dsurefire.rerunFailingTestsCount=2 -Dfailsafe.rerunFailingTestsCount=2" >> $GITHUB_ENV`.
Should we edit tests or build job?

Croway:
> Tests have failed because we run `@DirtiesContext` tests in parallel: `echo "MVND_OPTS=-V -e -ntp -Dmvnd.threads=2 -Daether.connector.http.connectionMaxTtl=120 -Daether.connector.requestTimeout=300000 -Daether.dependencyCollector.impl=bf -Dmaven.artifact.threads=25 -Dci.env.name=github.com -Dsurefire.rerunFailingTestsCount=2 -Dfailsafe.rerunFailingTestsCount=2" >> $GITHUB_ENV`. Should we edit tests or build job?

is the failure related to `Failed to start route myget because of duplicate id detected: myget`? In case can you use UUIDs or test class/name?

kulagaIA:
> 
> is the failure related to `Failed to start route myget because of duplicate id detected: myget`? In case can you use UUIDs or test class/name?

Changed tests to have unique route id's

PR #1219, opened 2024-09-13
Decrypt properties earlier in the SB lifecycle

Croway:
@luigidemasi

Using `camel.component.jasypt.early-decryption-enabled=true` the encryption of the properties will happen early in the life cycle, before Spring Boot context creation. it can be useful to resolve Spring Boot Auto Configuration properties, like ssl passwords and so on.


oscerd (approved):

davsclaus (approved):

davsclaus:
A JIRA ticket would be good so we can track this in the release notes

PR #1244, opened 2024-10-04
Use SB Async Interceptor

Croway:
The previous implementation was not using full Async code therefore we noticed issues under some circumstances.

The following implementation uses the Async interceptor provided by Spring Boot, right now the stack trace is similar to a plain Spring Boot Async RestController.

### DIFF
--- a/components-starter/camel-platform-http-starter/pom.xml
+++ b/components-starter/camel-platform-http-starter/pom.xml
@@ -61,6 +61,10 @@
       <version>${jetty-version}</version>
       <scope>test</scope>
     </dependency>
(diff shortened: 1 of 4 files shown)


oscerd (approved):

davsclaus (approved):

Croway:
It seems like that the PR failure is due to a connection refused against apache repository, maybe it was down temporarly?

`Connect to repository.apache.org:443 [repository.apache.org/65.109.119.155] failed: Connection refused`

davsclaus:
no problem you have tested it locally - okay to merge

PR #1260, opened 2024-10-18
Add camel-kamelets-catalog and camel-kamelets-utils to the camel-spring-boot BOM

cunningt:
@oscerd

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,6 +112,7 @@
 
         <!-- Camel target version -->
         <camel-version>4.9.0-SNAPSHOT</camel-version>
+        <camel-kamelets-version>4.9.0-SNAPSHOT</camel-kamelets-version>
 
         <!-- versions -->
         <aether-version>1.0.2.v20150114</aether-version>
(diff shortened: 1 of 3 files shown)


oscerd (changes requested):
The problem is we are going to release camel and csb first. So the released bom won't be aligned. I don't think there is a good solution for this unless we release kamelets after camel and then csb.

rinaldodev:
> The problem is we are going to release camel and csb first. So the released bom won't be aligned. I don't think there is a good solution for this unless we release kamelets after camel and then csb.

camel -> kamelets -> csb does seem like a good approach, but does it affect this PR?

oscerd:
I'm against changing the release order. It is working fine upstream and adding another step is not on the roadmap. Today camel and csb are released together in a unique vote. Adding Kamelets complicates things

davsclaus (changes requested):
same as andrea

rinaldodev:
I see. So if I understood correctly we are not including the catalog as a dependency in neither camel or csb, right?

If true, then I believe we should update these instructions? https://github.com/apache/camel-kamelets/blob/4.8.x/README.md?plain=1#L5

oscerd:
We are including camel kamelets dependency in camel jbang

Il sab 19 ott 2024, 17:31 Rinaldo Pitzer Júnior ***@***.***>
ha scritto:

PR #1268, opened 2024-10-24
CAMEL-21387: camel-opentelemetry-starter add default dependencies for…

johnpoth:
… basic setup

https://issues.apache.org/jira/browse/CAMEL-21387

I think it would be cool to add some docs for this. What would be the best way to do that ? Thanks !

### DIFF
--- a/components-starter/camel-opentelemetry-starter/pom.xml
+++ b/components-starter/camel-opentelemetry-starter/pom.xml
@@ -39,6 +39,15 @@
       <artifactId>camel-opentelemetry</artifactId>
       <version>${camel-version}</version>
     </dependency>
+    <dependency>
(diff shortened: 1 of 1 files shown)


Croway:
There is already a section in camel-opentelemtry docs, I think that you can add more documentation in https://github.com/apache/camel/blob/main/components/camel-opentelemetry/src/main/docs/opentelemetry.adoc#spring-boot

Thanks!

davsclaus (approved):

davsclaus:
We need to remember to remove the bridge if/when SB drops micrometer

johnpoth:
@davsclaus thanks will do (didn't know about that)

PR #1272, opened 2024-10-29
CAMEL-21395: camel-debezium - Use dedicated package for each db type - v 4.8

essobedo:
Fixes https://issues.apache.org/jira/browse/CAMEL-21395 (v 4.8)

## Motivation

The project camel-debezium-common shares the same package with the other camel-debezium sub-projects, which causes issues on environments like OSGI.

## Modifications:

* Use a dedicated package for each db type


essobedo:
Depends on https://github.com/apache/camel/pull/16114

PR #1282, opened 2024-11-15
feat(components): add observability services starter

squakez:
In this PR we're adding the starter for the component.

### DIFF
--- a/components-starter/camel-observability-services-starter/src/main/resources/config/application.properties
+++ b/components-starter/camel-observability-services-starter/src/main/resources/config/application.properties
@@ -1,0 +1,21 @@
+
+## ---------------------------------------------------------------------------
+## Licensed to the Apache Software Foundation (ASF) under one or more
+## contributor license agreements.  See the NOTICE file distributed with
+## this work for additional information regarding copyright ownership.
+## The ASF licenses this file to You under the Apache License, Version 2.0
+## (the "License"); you may not use this file except in compliance with
+## the License.  You may obtain a copy of the License at
+##
+##      http://www.apache.org/licenses/LICENSE-2.0
+##
+## Unless required by applicable law or agreed to in writing, software
+## distributed under the License is distributed on an "AS IS" BASIS,
+## WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+## See the License for the specific language governing permissions and
+## limitations under the License.
+## ---------------------------------------------------------------------------
+
+management.endpoints.web.exposure.include=prometheus
+management.endpoints.web.base-path=/observe
+management.endpoints.web.path-mapping.prometheus=metrics
--- a/components-starter/camel-observability-services-starter/pom.xml
+++ b/components-starter/camel-observability-services-starter/pom.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one or more
+    contributor license agreements.  See the NOTICE file distributed with
+    this work for additional information regarding copyright ownership.
+    The ASF licenses this file to You under the Apache License, Version 2.0
+    (the "License"); you may not use this file except in compliance with
+    the License.  You may obtain a copy of the License at
+
+         http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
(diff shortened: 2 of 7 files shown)


components-starter/camel-observability-services-starter/src/main/resources/config/application.properties
```diff
@@ -0,0 +1,21 @@
...
+##
+## Unless required by applicable law or agreed to in writing, software
+## distributed under the License is distributed on an "AS IS" BASIS,
+## WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+## See the License for the specific language governing permissions and
+## limitations under the License.
+## ---------------------------------------------------------------------------
+
+management.endpoints.web.exposure.include=prometheus
+management.endpoints.web.base-path=/camel
```
Croway, line 20: Is this default value correct? I may be mistaken here, but what if there are other components (platform-http for example) that uses `/camel` as base path?

davsclaus: Ah yeah that is correct using `/camel` is not the best idea
We could use a new term to avoid clashes `/camel-observability` - its a bit long to type, but SB uses `/actuator` and quarkus uses `/q`.

Croway: ah, do I understand correctly, the base-path refer to the actuators path? for example localhost:8080/actuators/camel?

squakez: Just tested the platform-http component and it does not affect its behavior:
```
$ curl http://localhost:8080/hello
Hello 
$ curl http://localhost:8080/camel/metrics
# HELP application_ready_time_seconds Time taken for the application to be ready to service requests
# TYPE application_ready_time_seconds gauge
application_ready_time_seconds{main_application_class="org.example.project.platformhttpserver.CamelApplication"} 3.979
....
```

Croway: IIRC there is one component that default to /camel, let me search that..

O yeah, it wasn't platform-http, but servlet https://camel.apache.org/components/4.8.x/servlet-component.html#_sb_option_camel_servlet_mapping_context-path

squakez: > Ah yeah that is correct using `/camel` is not the best idea We could use a new term to avoid clashes `/camel-observability` - its a bit long to type, but SB uses `/actuator` and quarkus uses `/q`.

That's the idea of this component, "to normalize" the endpoints across the different runtimes. Consider that all the actuators should be eventually exposed with this normalized endpoints (in all runtimes). If the user does not want to do that, then, it should not use this component at all IMO.

Croway: One last question, what's the difference using /camel/metrics and the Spring Boot actuators /actuator/metrics?

squakez: > ah, do I understand correctly, the base-path refer to the actuators path? for example localhost:8080/actuators/camel?

The base path would only refer to the actuators exposed. So, instead of `/actuator/health` we'll have `/camel/health`. However, as I said, the idea is to bring into this component all the different endpoints, so, in theory, when the user is using this, he should know that the expected endpoints are published under this normalize url. Of course, I would not expect the user to use this component + other components which are exposing actuators (say, health).

squakez: if we want to avoid any possible clash with `camel` (which may be widely used as a base path by default) we can call it `observe` or `metrics` maybe.

squakez: Some more test to verify possibility to clash when user specify `camel` as a platform http uri.
```
$ curl http://localhost:8080/camel/hello
Hello
$ curl http://localhost:8080/camel/metrics
# HELP application_ready_time_seconds Time taken for the application to be ready to service requests
# TYPE application_ready_time_seconds gauge
application_ready_time_seconds{main_application_class="org.example.project.platformhttpserver.CamelApplication"} 3.988
```
It does not interfere apparently.

squakez:
@davsclaus @Croway I've tested the component with a servlet, and the clash, although did not produce any error, really was not allowing the exposure of the metrics. It makes sense to move the base default to `/observe/`, hence having `/observe/metrics`, `/observe/health`, ... I'm changing it also in the core component.

davsclaus:
Yes `observe` is a good name

And about servlet then it was before we had platform-http on SB. So its not recommended to use. But IMHO lets avoid using camel as end users may use that too.

Croway (approved):

squakez:
Default changed also in core: https://github.com/apache/camel/pull/16302

PR #1298, opened 2024-11-29
Fix platform-http reload issues

Croway:


davsclaus (approved):

davsclaus:
I wonder if SB calls `unregisterHttpEndpoint` on shutdown. This is empty currently. But just wondered if its in use at all or not.

oscerd (approved):

PR #1299, opened 2024-11-29
CAMEL-21461: platform-http - add CORS support

rinaldodev:
Enable CORS support when using platform-http.

I'm unsure if this caching [here](https://github.com/rinaldodev/camel-spring-boot/blob/951244227f5531b2ae4301e0a27907e252e51a34/components-starter/camel-platform-http-starter/src/main/java/org/apache/camel/component/platform/http/springboot/CamelRequestHandlerMapping.java#L86-L90) could be an issue. Do we allow changing CORS configurations after initialization?

```java
  if (corsConfiguration == null) {
      Map<String, String> corsHeaders = restConfiguration.getCorsHeaders();
      corsConfiguration = createCorsConfiguration(corsHeaders != null ? corsHeaders : Collections.emptyMap());
  }
  return corsConfiguration;
```

### DIFF
--- a/components-starter/camel-platform-http-starter/src/test/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpCorsCredentialsTest.java
+++ b/components-starter/camel-platform-http-starter/src/test/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpCorsCredentialsTest.java
@@ -22,7 +22,6 @@ import org.apache.camel.spi.RestConfiguration;
 import org.apache.camel.spring.boot.CamelAutoConfiguration;
 import org.apache.camel.test.spring.junit5.CamelSpringBootTest;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.security.oauth2.client.servlet.OAuth2ClientAutoConfiguration;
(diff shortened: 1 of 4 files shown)


components-starter/camel-platform-http-starter/src/test/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpCorsCredentialsTest.java
```diff
@@ -0,0 +1,108 @@
...
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.*;
+
+@EnableAutoConfiguration(exclude = {OAuth2ClientAutoConfiguration.class, SecurityAutoConfiguration.class})
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+@CamelSpringBootTest
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = { CamelAutoConfiguration.class,
+        SpringBootPlatformHttpCorsCredentialsTest.class, SpringBootPlatformHttpCorsCredentialsTest.TestConfiguration.class,
+        PlatformHttpComponentAutoConfiguration.class, SpringBootPlatformHttpAutoConfiguration.class, })
+public class SpringBootPlatformHttpCorsCredentialsTest {
```
Croway, line 28: can you add a negative test please?
ala
```given()
            .when()
            .get("/rest")
            .then()
            .statusCode(4xx/5xx)
            ```

Croway:
@rinaldodev `The following syntax doesn't work when testing CORS because it seems to overrides the behavior of all HTTP methods including OPTIONS` this sounds odd, what about using `httpMethodRestrict`? do you think this can help somehow?

davsclaus (commented):
Just a small thing, java source files should have ASF license headers.

components-starter/camel-platform-http-starter/src/test/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpCorsCredentialsTest.java
rinaldodev: I'm looking into it.

davsclaus:
Thanks @rinaldodev more features for CSB in platform-http is welcome.

PR #1326, opened 2024-12-23
Align Jakarta Jaxb 4.0.2 with Camel core

apupier:
I'm wondering why it is not inheriting the version from Camel core bom. And if there is a good reason, I guess we should add it to the allow list of dependabot to have automatic updates


davsclaus:
There is no specific reason for not aligning to core

oscerd (approved):

PR #1355, opened 2025-01-28
CAMEL-21673 Fix missing stream closes; fix getBackOffMaxDelay usage

cunningt:
https://issues.apache.org/jira/browse/CAMEL-21673

Fix several missing stream closes; fix a getBackOffMaxDelay reference that incorrectly references getBackDelay; refresh


oscerd (approved):

components-starter/camel-platform-http-starter/src/main/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpBinding.java
```diff
@@ -168,6 +168,7 @@ protected void doWriteDirectResponse(Message message, HttpServletResponse respon
                     } finally {
                         IOHelper.close(new Closeable[]{is, os});
                     }
+                    stream.close();
```
davsclaus, line 171: Can you remove this as we should not close the http stream at this point

davsclaus (changes requested):

components-starter/camel-platform-http-starter/src/main/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpBinding.java
davsclaus: Ah sorry my bad, you can likely move stream into the finally block as above

IOHelper.close(new Closeable[]{is, os, stream});

davsclaus (approved):

components-starter/camel-platform-http-starter/src/main/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpBinding.java
cunningt: Can definitely remove it but this dumb tool is going to keep popping the error on stream not being closed out - where is the right place to close it?    Do we need to use IOHelper?

davsclaus: IOHelper ignores closing exceptions, in case .close fails

cunningt: @davsclaus Really sorry, but I'm still confused here about why the stream variable can't be closed out at the end of the if block because it looks like it goes out of scope there - is there somewhere we can close it out so we don't have to deal with this Coverity test failing?

cunningt: Sorry - missed your comment above about the finally block!

PR #1357, opened 2025-01-30
feat(components/micrometer): baseEndpointUriExchangeEventNotifier parameter

squakez:
ref CAMEL-21654


oscerd (approved):

squakez:
I think the failure is because the change expected is not yet in the Apache Snapshot repo. Let's retry the check tomorrow.

davsclaus (approved):

PR #1358, opened 2025-01-30
chore: regen sync

squakez:
Building on `main` regenerated these files.


oscerd (changes requested):
This is not aligned with the 4.10.0-SNAPSHOT upstream, because the header is there already. You should rebuild the snapshot fully.

squakez:
I'm closing this. I'll open in case I find any other change.

PR #1362, opened 2025-02-04
CAMEL-21701: Make single file upload easy to access from message body

Croway:
The behavior is slightly different compared to plain Camel https://github.com/apache/camel/pull/17045

the `attachment.getContent()` in Camel is a String, while, in Camel Spring Boot, a FileInputStream, therefore https://github.com/apache/camel-spring-boot/compare/main...Croway:CAMEL-21701?expand=1#diff-e11e7487d45a9ac2f3cc2803af2694c0d238bc6720d49b1cb9e8f1ea4494a99bR109


davsclaus (approved):

davsclaus:
yeah `getContent` dependends on the implementation - it is usually an InputStream. There is also getInputStream API I think you can use instead of getContent to be specific it should be stream

PR #1363, opened 2025-02-04
CAMEL-21544: Add option to turn on|off whether application/x-www-form…

Croway:
…-urlencoded should populate Map body with form

Again, slightly different behavior from main/quarkus. In Camel Spring Boot the body always contains a Stream, the difference is that when `populateBodyWithForm=true` (which is the default) the body is a Map, not a Stream.


Croway:
The build failure is expected, the SNAPSHOT has to be built with https://github.com/apache/camel/pull/17048

davsclaus (approved):

oscerd (approved):

PR #1372, opened 2025-02-18
feat(components): onboard camel-telemetry-dev

squakez:


components-starter/camel-telemetry-dev-starter/src/main/java/org/apache/camel/telemetrydev/starter/TelemetryDevAutoConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.camel.telemetrydev.TelemetryDevTracer;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration(proxyBeanMethods = false)
+@EnableConfigurationProperties(TelemetryDevConfigurationProperties.class)
+@ConditionalOnProperty(value = "camel.telemetryDev.enabled", matchIfMissing = true)
```
Croway, line 29: I do not know if this can be an issue, but, here the property is `telemetryDev` (camel case), while in https://github.com/apache/camel-spring-boot/pull/1372/files#diff-50a0b75a5edcad1102b5bb774ce1a543736102057e656ce19b1ce545451cd951R21 is `telemetrydev`

squakez: Good spotted. Yes, I originally used the snake case as we have in the core, but later the runtime complained as it required to be kebab case instead in SB. I'm going to change it before the merge, thanks!

Croway (approved):

PR #1381, opened 2025-02-25
Fix deprecated methods and classes

Croway:
Hi @gnodet I made some changes to Mojos, it would be nice if you can take a look, moreover, `SpringBootStarterMojo` uses `ProjectDependenciesResolver` and `project.getProjectBuildingRequest()` that are deprecated, without a replacement.


oscerd (approved):

davsclaus:
@gnodet can you take a quick look - in those mojo changes. So it seems like can you use `@Inject` now to inject instead of an old maven annotation

gnodet (approved):
LGTM

gnodet:
> @gnodet can you take a quick look - in those mojo changes. So it seems like can you use `@Inject` now to inject instead of an old maven annotation

Yes, that's right, there's no good reason to use `@Component` anymore.

PR #1473, opened 2025-07-29
CAMEL-22282: Test for platform-http bridged endpoint

mcarlett:
test for https://github.com/apache/camel/pull/18755


mcarlett:
the failure on the test is expected since it needs the PR on Camel to make it pass

davsclaus (approved):

PR #1491, opened 2025-08-26
feat(starters): onboard Micrometer Observability starter

squakez:
Ref CAMEL-21869


squakez:
We may need to wait https://github.com/apache/camel/pull/18997 ends up in the snapshot

PR #1495, opened 2025-09-01
Make Virtual threads more integrated with Camel Spring Boot

Croway:


davsclaus (approved):

davsclaus:
the spring factory file may need an ASF license header

PR #1517, opened 2025-09-22
CAMEL-22437 - Create Docling Spring Boot Starter

oscerd:


oscerd:
Merged because Docling component is not in the snapshot yet.

PR #1709, opened 2026-03-12
CAMEL-22760: Bridge spring.kafka.* properties to camel-kafka component

gnodet:
## Summary

When using `camel-kafka-starter` with Spring Boot, users previously had to duplicate their Kafka configuration under both `spring.kafka.*` and `camel.component.kafka.*` properties. This PR adds automatic bridging so that Spring Boot's standard Kafka properties are picked up by the Camel Kafka component.

### Properties bridged

| Spring Boot (`spring.kafka.*`) | Camel (`camel.component.kafka.*`) |
|---|---|
| `bootstrap-servers` | `brokers` |
| `client-id` | `client-id` |
| `security.protocol` | `security-protocol` |
| `consumer.group-id` | `group-id` |
| `ssl.key-store-location` | `ssl-keystore-location` |
| `ssl.key-store-password` | `ssl-keystore-password` |
| `ssl.key-store-type` | `ssl-keystore-type` |
| `ssl.key-password` | `ssl-key-password` |
| `ssl.trust-store-location` | `ssl-truststore-location` |
| `ssl.trust-store-password` | `ssl-truststore-password` |
| `ssl.trust-store-type` | `ssl-truststore-type` |
| `ssl.protocol` | `ssl-protocol` |
| `properties[sasl.mechanism]` | `sasl-mechanism` |
| `properties[sasl.jaas.config]` | `sasl-jaas-config` |
| `properties[sasl.kerberos.service.name]` | `sasl-kerberos-service-name` |

### Design

- Uses Spring Boot's `Binder` to detect which `camel.component.kafka.*` properties the user explicitly set
- Only bridges a Spring Boot value when the user has NOT explicitly set the corresponding Camel property
- Explicit `camel.component.kafka.*` settings always take precedence
- `@AutoConfigureBefore(KafkaComponentAutoConfiguration.class)` ensures bridging happens before the generated auto-configuration copies properties to the component
- `@ConditionalOnClass(KafkaProperties.class)` ensures it only activates when `spring-boot-kafka` is on the classpath (optional dependency)

## Test plan

- [x] 9 unit tests covering all bridged properties, precedence, and edge cases
- [x] All tests pass
- [ ] CI build passes
- [ ] Integration test with a real Kafka broker (manual)

### DIFF
--- a/components-starter/camel-kafka-starter/pom.xml
+++ b/components-starter/camel-kafka-starter/pom.xml
@@ -52,6 +52,12 @@
       </exclusions>
       -->
     </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-kafka</artifactId>
+      <version>${spring-boot-version}</version>
+      <optional>true</optional>
+    </dependency>
     <!-- test -->
     <dependency>
       <groupId>org.apache.camel</groupId>
--- a/components-starter/camel-kafka-starter/src/main/java/org/apache/camel/component/kafka/springboot/SpringKafkaPropertiesAutoConfiguration.java
+++ b/components-starter/camel-kafka-starter/src/main/java/org/apache/camel/component/kafka/springboot/SpringKafkaPropertiesAutoConfiguration.java
@@ -25,6 +25,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.bind.Bindable;
 import org.springframework.boot.context.properties.bind.Binder;
 import org.springframework.boot.kafka.autoconfigure.KafkaAutoConfiguration;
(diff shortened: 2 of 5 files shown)


gnodet:
Added the toggle in the latest commit. The bridge can be disabled with:

```properties
camel.component.kafka.bridge-spring-kafka-properties=false
```

It's enabled by default (`matchIfMissing = true`). Uses `@ConditionalOnProperty` so the entire auto-configuration class is skipped when disabled — no beans created, no overhead.

davsclaus (approved):

davidkarlsen:
What about `transactional()` - https://camel.apache.org/components/4.18.x/eips/transactional-client.html - spring has a KafkaTransactionManager (PlatformTransactionManager) - or would that be too complex?
Perhaps just add a note in https://camel.apache.org/components/4.18.x/kafka-component.html#_kafka_transaction about how it is intended to work with tx.

davsclaus:
@davidkarlsen can you create a JIRA about Spring Kafka Transaction support for camel-kafka

davsclaus:
> On/off toggle: Good point — we could add a camel.component.kafka.bridge-spring-kafka-properties (or similar) property. We can add that in a follow-up if you'd like, or in this PR. What do you prefer?

Yes this PR

> Migration guide note: Will add a note to the 4.19 migration guide.

Add this to the 4.21 migration guide (4.19 and 4.20 has been released)

@gnodet ^^^^

davsclaus:
@Croway have you seen this PR.

The spring-boot kafka starter is added as optional=true dependency. Which would be needed for this PR. Though the con could be that we have a Camel vs SB kafka client misalignment and a classpath problem. But that can happen anyway today. And I guess if I recall correct optional=true then the SB kafka JARs are not loaded by default by maven, and wont be in the end user app.

Croway:
@davsclaus exactly, the PR is fine, I was even thinking to remove the optional=true, and always import `spring-boot-kafka`, but this will add Spring Boot `KafkaTemplate` and Kafka configurations classes even if the user is not using it. Therefore, this is the right approach.

I just tested the version alignments and I confirm that the Kafka client version is aligned by the Spring Boot BOM. For example, camel 4.18.1 declares kafka 3.9.x. The Spring Boot version used in Camel Spring Boot 4.18.1, declares kafka 3.8.x, that is the version that camel spring boot users uses at runtime.

I'll create an issue to harden the tests and CI so that we might be able to catch potential API changes and misalignments.

Croway (approved):

davsclaus:
@gnodet or @Croway can any of you help update the 4.21 upgrade guide (in camel-core) about this - as CSB users that use spring-boot-kafka would be good to be aware of this feature when they are upgrading.

Croway:
will do, I'll provide an entry in the 4.21 upgrade guide.

EDIT: https://github.com/apache/camel/pull/22786

PR #1745, opened 2026-04-11
Fix CI parse_errors.sh and replace deprecated AvailablePortFinder

cunningt:
This commit was started to replace some deprecated API used in tests but kind of morphed because there were a lot of unrelated issues with running the incremental tests in CI.      Here are the changes

- replace AvailablePortFinder.getNextAvailable() (deprecated) usage in a few tests
- add a parse_errors.sh script - this was missing in camel-spring-boot and incremental tests were failing because it didn't exist
- change CamelVirtualThreadsTests to an integration test : I tried a lot of avenues to try to get this test to work in CI, but it appears that the thread type gets set by the first test run and we were getting platform threads within the test every time in CI.   I've changed it to be an integration test that runs if we're not in CI (there's a profile that checks for the ci.env name) because even after changing this to an IT it was failing due to previous tests run in surefire.

### DIFF
--- a/core/camel-spring-boot/src/test/java/org/apache/camel/spring/boot/CamelVirtualThreadsIT.java
+++ b/core/camel-spring-boot/src/test/java/org/apache/camel/spring/boot/CamelVirtualThreadsIT.java
@@ -29,6 +29,8 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledForJreRange;
 import org.junit.jupiter.api.condition.JRE;
 import org.junit.jupiter.api.parallel.Isolated;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.test.context.SpringBootTest;
@@ -54,6 +56,8 @@ import static org.assertj.core.api.Assertions.assertThat;
                     "camel.threads.virtual.enabled=true"
                 })
 public class CamelVirtualThreadsIT {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CamelVirtualThreadsIT.class);
 
     @Autowired
     CamelContext context;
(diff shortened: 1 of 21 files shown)


oscerd (approved):

Croway (approved):

core/camel-spring-boot/src/test/java/org/apache/camel/spring/boot/CamelVirtualThreadsIT.java
```diff
@@ -97,9 +114,27 @@ public void testCamelVirtualThreadPropertyIsSet() throws Exception {
     @Test
     @EnabledForJreRange(min = JRE.JAVA_21)
     public void testRouteExecutesOnVirtualThread() throws Exception {
+        // Debug: Print configuration at test start
+        System.err.println("=== testRouteExecutesOnVirtualThread Debug ===");
```
Croway: can LOG be used instead?

core/camel-spring-boot/src/test/java/org/apache/camel/spring/boot/CamelVirtualThreadsIT.java
```diff
@@ -89,6 +98,14 @@ private static boolean isVirtualThread(Thread thread) {
     public void testCamelVirtualThreadPropertyIsSet() throws Exception {
         // Verify that the environment post processor set the camel.threads.virtual.enabled property
         String camelVirtualThreadsProperty = System.getProperty("camel.threads.virtual.enabled");
+
+        // Debug output to understand what's happening
+        System.out.println("=== Virtual Thread Configuration Debug ===");
```
Croway: can LOG be used instead?

core/camel-spring-boot/src/test/java/org/apache/camel/spring/boot/CamelVirtualThreadsIT.java
cunningt: Sorry about that!   That was leftover from debugging, changed to LOG and pared down the output.

core/camel-spring-boot/src/test/java/org/apache/camel/spring/boot/CamelVirtualThreadsIT.java
cunningt: Sorry about that! That was leftover from debugging, changed to LOG and pared down the output.

Croway (approved):

PR #1771, opened 2026-04-29
CAMEL-23378: avoid String to InputStream conversion

mcarlett:
Spring's ObjectToObjectConverter finds FileInputStream(String) constructor and treats the String value as a file path instead of data content.


davsclaus (approved):

Croway (approved):

davsclaus:
i have backported to 4.18 LTS

PR #1778, opened 2026-05-04
Do not overwrite async timeout 503 with sendError(500) in afterProcess

Croway:
## Summary

- When Spring's async timeout handler sends 503, the Camel executor thread's `afterProcess()` catches `AsyncRequestNotUsableException` from `writeResponse()` and falls through to `sendError(500)`, overwriting the 503
- Detect `AsyncRequestNotUsableException` in `afterProcess()` and `service()` and skip the `sendError(500)` fallback
- Uses class name check to avoid hard compile-time dependency on `spring-web`

## Test plan

- [x] New `SpringBootPlatformHttpAsyncNotUsableTest` with two test methods:
  - `testAsyncNotUsableShouldNotOverwriteWith500` — verifies no 500 when `AsyncRequestNotUsableException` is thrown
  - `testRegularIOExceptionStillReturns500` — verifies regular `IOException` still produces 500
- [x] Test fails before fix (gets 500), passes after fix
- [x] All 99 platform-http tests pass (0 failures, 0 errors) when combined with #1777


Croway:
The failures are expected, https://github.com/apache/camel-spring-boot/pull/1777 this PR fixes those failing tests.

PR #1780, opened 2026-05-08
feat(starter): enable telemetry disable processor parameter

squakez:
Ref https://github.com/apache/camel/pull/23002


squakez:
Checks failure expected. We need to wait camel main from upstream to be available here.

davsclaus (approved):

PR #1847, opened 2026-07-12
CAMEL-24025: ship observability-services defaults via EnvironmentPostProcessor

Croway:
Resolves [CAMEL-24025](https://issues.apache.org/jira/browse/CAMEL-24025)

## What

`camel-observability-services-starter` packaged its opinionated defaults as a literal `config/application.properties` inside the starter jar. Spring Boot's config-data search gives `classpath:/config/application.properties` higher precedence than the application's own `classpath:/application.properties`, so the starter defaults silently overrode user settings — customization was only possible via environment variables or system properties, and this limitation was even documented in the component docs (referencing spring-projects/spring-boot#24688).

This PR replaces the packaged properties file with an `ObservabilityServicesEnvironmentPostProcessor` registered via `META-INF/spring.factories`. It contributes the exact same defaults as a `MapPropertySource` added with `addLast(...)` (running at `Ordered.LOWEST_PRECEDENCE`, i.e. after `ConfigDataEnvironmentPostProcessor`), so the defaults sit at the bottom of the environment and every user-provided source overrides them following standard Spring Boot precedence rules. This aligns the behaviour with the Camel Main and Quarkus variants of `camel-observability-services`.

## Behavioural change

Previously the packaged defaults **won** over the user's `application.properties`; now the user's configuration wins. This is the intent of the ticket, but applications that set e.g. `management.server.port` and silently got `9876` anyway will now get their own value.

## Tests

New `ObservabilityServicesEnvironmentPostProcessorTest`:
- `defaultsAreApplied` — all defaults resolve from the environment when not overridden
- `userApplicationPropertiesOverrideDefaults` — values from a regular `application.properties` win over the starter defaults (this assertion fails against the previous packaged-file mechanism)

## Follow-up

The WARNING in the `camel-observability-services` docs (apache/camel, `observability-services.adoc`) stating that customization is not available on Spring Boot becomes obsolete once this is merged — will follow up with a docs PR in apache/camel.

---
_This PR was generated by Claude Code on behalf of Federico Mariani_

### DIFF
--- a/components-starter/camel-observability-services-starter/pom.xml
+++ b/components-starter/camel-observability-services-starter/pom.xml
@@ -60,6 +60,13 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <!-- test dependencies -->
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-test</artifactId>
+      <version>${spring-boot-version}</version>
+      <scope>test</scope>
+    </dependency>
     <!--START OF GENERATED CODE-->
     <dependency>
       <groupId>org.apache.camel.springboot</groupId>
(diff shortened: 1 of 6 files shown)


davsclaus (approved):
Great to spot this so we have good usability on SB

squakez:
Nice one @Croway . I think this is also fixing this behavior: https://camel.apache.org/components/next/others/observability-services.html#_configuration - if so, feel free to remove the note explaining that the component is not configurable in SB.

squakez:
> Nice one @Croway . I think this is also fixing this behavior: https://camel.apache.org/components/next/others/observability-services.html#_configuration - if so, feel free to remove the note explaining that the component is not configurable in SB.

Nevermind, I've seen you've done this already!

PR #1859, opened 2026-07-20
fix: NPE when a Spring Boot config source resolves to an empty YAML object

vilmosnagy:
Previously, Camel could start successfully when a loaded YAML file's root object was empty (`{}`). With Spring Boot's recent changes (https://github.com/spring-projects/spring-boot/issues/48958), when the root object is empty (`{}`), the resulting `EnumerablePropertySource` now contains a `null -> ""` mapping (previously this property source was empty). Because of this, every call to `eps.getPropertyNames()` must now guard against a null name.

Confirmed this is a regression: without this fix, the new test throws an NPE against the current Spring Boot version. The same test passes against Camel 4.20 (pre-upgrade), which used an older Spring Boot version where this property source was empty rather than containing a null key.


davsclaus (approved):

davsclaus:
Thanks for the PR

vilmosnagy:
thanks for the quick merge :slightly_smiling_face:

PR #1874, opened 2026-08-04
CAMEL-24351: camel-platform-http-starter - the request mapping must not be lazy

henrik242:
`@Lazy` was added in 4.21.0, replacing an ObjectProvider that deferred the CamelContext lookup. CamelRequestHandlerMapping registers itself as a PlatformHttpListener and is only notified of endpoints created after it exists, so creating it on first demand is unordered with respect to CamelContext startup. When Camel starts first the mapping learns about no endpoint at all, and every platform-http route answers 404 while reporting itself as started. 4.18.3 and 4.20.0 both instantiated the bean eagerly.

Issue link: CAMEL-24351

### DIFF
--- a/components-starter/camel-platform-http-starter/src/test/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpEagerMappingTest.java
+++ b/components-starter/camel-platform-http-starter/src/test/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpEagerMappingTest.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.camel.component.platform.http.springboot;
+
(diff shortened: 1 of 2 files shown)


Croway (approved):

components-starter/camel-platform-http-starter/src/test/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpEagerMappingTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.servlet.mvc.method.RequestMappingInfo;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
```
davsclaus: Minor: Camel prefers AssertJ assertions over JUnit. Consider:
```suggestion
import static org.assertj.core.api.Assertions.assertThat;
```
and then `assertThat(...).isFalse()` / `assertThat(...).isTrue()` instead of `assertFalse`/`assertTrue`. Non-blocking.

davsclaus (approved):
Good fix for a real user-facing bug — the `@Lazy` on `CamelRequestHandlerMapping` caused silent 404s because the mapping registers as a `PlatformHttpListener` in its constructor and only receives notifications for endpoints created *after* it exists. When lazy, Camel starts first, registers all endpoints, and the mapping never learns about any of them.

The `ObjectProvider<CamelContext>` approach is the standard Spring idiom for breaking the circular dependency that motivated `@Lazy` in the first place. The Javadoc clearly explains the design rationale, and the test covers both the eager-instantiation invariant and actual endpoint mapping.

**Minor (non-blocking):** the test uses JUnit `assertFalse`/`assertTrue` — Camel prefers AssertJ (`assertThat(...).isFalse()`).

Note: this review covers project conventions and rules compliance. It does not replace specialized AI review tools (CodeRabbit, Sourcery) or static analyzers (SonarCloud).

_Claude Code on behalf of @davsclaus_

PR #1897, opened 2026-08-26
CAMEL-24496: camel-platform-http-starter - align SpringBootPlatformHttpBinding multipart handling

oscerd:
Aligns `SpringBootPlatformHttpBinding.populateAttachments()` with the equivalent bindings in the other
HTTP server components. Three items, all in that one method.

### 1. `fileNameExtWhitelist` was evaluated against the multipart field name

`getFileMap().forEach((name, multipartFile) -> ...)` gives `name` = the multipart **field** key, not the
submitted file name — but line 135 propagated `multipartFile.getOriginalFilename()` downstream regardless.
With `fileNameExtWhitelist=pdf`, a part with field name `file` carrying `filename="shell.jsp"` was accepted,
because `FileUtil.onlyExt("file")` is `null` and the check was skipped.

Two further problems in the same block:

- **Fail-open on a missing extension.** `if (ext != null)` left `accepted = true` when the name had no
  extension, so a whitelist never rejected an extension-less name.
- **Substring matching.** `getFileNameExtWhitelist().contains(ext)` meant a whitelist of `pdf` also matched
  `invoice.pd`.

The check now runs against `getOriginalFilename()`, fails closed when a whitelist is configured and the name
has no extension, and compares whole comma-separated tokens case-insensitively. `FileUtil.onlyExt`'s
first-dot semantics are preserved so a whitelist entry such as `tar.gz` keeps working.

### 2. Rejected uploads were still written to the servlet temp directory

`transferTo()` ran before the whitelist decision and the rejection branch only logged, so a rejected upload
stayed on disk for the lifetime of the process. The decision now happens first — `getOriginalFilename()` is
available without transferring — so a rejected file never reaches disk.

### 3. `CamelFileName` set from the raw submitted name

`Exchange.FILE_NAME` is a control header consumed by the file/ftp producers. CAMEL-24293 applied
`FileUtil.stripPath(...)` to externally-derived names in `camel-platform-http-vertx`, `camel-zipfile` and
`camel-tarfile`; this binding lives in camel-spring-boot and was not covered by that change. Same
normalisation applied here, with the same comment.

`Exchange.FILE_PATH` and `FILE_LENGTH` are left as they are — they are deliberate (added in
`2c0d67f9f87`) and the vertx reference implementation sets `FILE_PATH` too.

### Tests

`SpringBootPlatformHttpFileNameExtWhitelistTest` — 7 cases. Verified meaningful: the 5 defect cases all fail
against the unpatched binding and pass with it; the 2 happy-path cases pass either way as regression guards.
Full module suite: 124 tests, 0 failures.

### Deliberately not in this PR

**Path variables from the raw URI** (also listed on CAMEL-24496). `populateRequestParameters()` evaluates
placeholders against `getRawPath()`, which returns the undecoded `request.getRequestURI()`. Changing that is
wider than it looks: the base-class `getRawPath()` also feeds `Exchange.HTTP_PATH` for every request, and the
override exists on purpose (CAMEL-22116, refined by CAMEL-23191 for context-path handling). Worth doing, but
it deserves its own change and its own discussion rather than riding along here.

### Note for CAMEL-24427

While checking the reference implementations: the **fail-open** and **substring** problems above are not
specific to this starter. `DefaultHttpBinding.populateAttachments()` in `camel-http-common` has both, and so
does `VertxPlatformHttpConsumer`. Only the wrong-value defect is camel-spring-boot's own. Since CAMEL-24427
is already open for camel-servlet/camel-jetty, those two may be worth folding into its scope so all the
engines agree — otherwise this binding is now stricter than the others.

---

### DIFF
--- a/components-starter/camel-platform-http-starter/src/main/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpBinding.java
+++ b/components-starter/camel-platform-http-starter/src/main/java/org/apache/camel/component/platform/http/springboot/SpringBootPlatformHttpBinding.java
@@ -104,50 +104,77 @@ public class SpringBootPlatformHttpBinding extends DefaultHttpBinding {
             message.setHeader(Exchange.ATTACHMENTS_SIZE, multipartHttpServletRequest.getFileMap().keySet().size());
             multipartHttpServletRequest.getFileMap().forEach((name, multipartFile) -> {
                 try {
-                    Path uploadedTmpFile = Paths.get(tmpFolder.getPath(), UUID.randomUUID().toString());
-                    multipartFile.transferTo(uploadedTmpFile);
-
                     if (name != null) {
                         name = name.replaceAll("[\n\r\t]", "_");
                     }
 
-                    boolean accepted = true;
-
-                    if (getFileNameExtWhitelist() != null) {
-                        String ext = FileUtil.onlyExt(name);
-                        if (ext != null) {
-                            ext = ext.toLowerCase(Locale.US);
-                            if (!getFileNameExtWhitelist().equals("*") && !getFileNameExtWhitelist().contains(ext)) {
-                                accepted = false;
-                            }
-                        }
-                    }
-
-                    if (accepted) {
-                        AttachmentMessage am = new DefaultAttachmentMessage(message);
-                        File uploadedFile = uploadedTmpFile.toFile();
-                        am.addAttachment(name, new DataHandler(new CamelFileDataSource(uploadedFile, name)));
-
-                        // populate body in case there is only one attachment
-                        if (isSingleAttachment) {
-                            message.setHeader(Exchange.FILE_PATH, uploadedFile.getAbsolutePath());
-                            message.setHeader(Exchange.FILE_LENGTH, multipartFile.getSize());
-                            message.setHeader(Exchange.FILE_NAME, multipartFile.getOriginalFilename());
-                            if (multipartFile.getContentType() != null) {
-                                message.setHeader(Exchange.FILE_CONTENT_TYPE, multipartFile.getContentType());
-                            }
-                            message.setBody(uploadedTmpFile);
-                        }
-                    } else {
+                    // the whitelist is evaluated against the submitted file name - the value that is
+                    // propagated downstream - and before the upload is written to the servlet temp
+                    // directory, so a rejected file never reaches disk
+                    if (!isFileNameExtAccepted(multipartFile.getOriginalFilename())) {
                         LOG.debug(
                                 "Cannot add file as attachment: {} because the file is not accepted according to fileNameExtWhitelist: {}",
                                 name, getFileNameExtWhitelist());
+                        return;
+                    }
+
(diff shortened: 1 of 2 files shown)


davsclaus (approved):
Reviewed against this project's build/test/contribution rules. I checked out the branch locally, built the module, ran the new test class, and cross-checked the history and precedent this PR relies on.

**Verification performed**
- Build: `mvn -DskipTests install` in `camel-platform-http-starter` succeeds.
- Tests: `SpringBootPlatformHttpFileNameExtWhitelistTest` — 7/7 pass.
- Root-cause claim (bug #1): confirmed via `git blame` / `git show cd535caf7af2` that a Feb 2025 commit switched the propagated `FILE_NAME` header to `multipartFile.getOriginalFilename()` but never updated the whitelist check a few lines above, which still ran against the multipart field key. This PR fixes a genuine, verifiable divergence rather than reverting intentional behavior.
- CAMEL-24293 precedent: verified against a local `apache/camel` checkout — `FileUtil.stripPath()` is applied the same way in `VertxPlatformHttpConsumer`, `TarFileDataFormat`/`TarIterator`, and `ZipFileDataFormat`/`ZipIterator` under that ticket. The pattern used here is consistent with established precedent.
- `FileUtil.onlyExt` semantics (first-dot, so `tar.gz` stays whole; null-safe) match what the new `isFileNameExtAccepted` helper assumes.

**Findings**

No blocking issues.

- Reordering the whitelist decision before `transferTo()` correctly prevents rejected uploads from touching disk — verified no leftover write calls in the rejected path.
- Whole-token, case-insensitive, trimmed whitelist matching correctly replaces the old substring/`contains` check (tested by `rejectsSubstringOfAWhitelistedExtension`, `acceptsSecondWhitelistEntryIgnoringSurroundingSpaces`).
- Fail-closed on a missing extension when a whitelist is configured — correct and tested.
- The new test class follows this module's established conventions exactly (rest-assured + permit-all `SecurityFilterChain`, same shape as `PlatformHttpBinaryUploadTest.java`).
- Single commit, correctly formatted (`CAMEL-24496: ...`) per project guidelines.
- No public API changes, no new dependencies, no Lombok/Records — compliant with project standards.
- Scope is tight and well justified; the PR explicitly and correctly defers the separate `getRawPath()` path-variable issue (also under CAMEL-24496) to its own change, with a clear explanation why.

**Minor, non-blocking observation**

`Exchange.ATTACHMENTS_SIZE` is still set to the raw multipart field count *before* whitelist filtering runs, so a caller can't currently distinguish "parts submitted" from "parts accepted." This is pre-existing behavior (unchanged by this PR — verified against the original code), not a regression. Not blocking; possibly worth a follow-up ticket if that distinction matters for consumers of this header.

oscerd:
Rebased onto `main` (force-update) now that #1905 has landed.

That PR removed six starters whose components were deleted from camel core, which was failing documentation validation on `main` — this branch was inheriting that failure through no fault of its own (`target of xref not found: next@components::ironmq-component.adoc`). With the rebase, `check-doc` should go green here.

No content change: the diff against `main` is byte-for-byte what it was before the rebase, and the module tests were re-run against the new `main` before pushing.

Croway (approved):

PR #1898, opened 2026-08-26
CAMEL-24499: camel-spring-boot - route detail view bypasses the route.start.exception serialization filter

oscerd:
`RouteEndpointInfo` annotates its properties map so `route.start.exception` is not serialized:

```java
@JsonIgnoreProperties(value = {"route.start.exception"})
private final Map<String, Object> properties;
```

`RouteDetailsEndpointInfo` extends it but re-declared the same field without the annotation, plus a
`getProperties()` override, so the detail operation serialised the property the base view filters.

Both the annotation and the shadowing field arrived in the same commit under CAMEL-20993 — the filter was
clearly intended, and the subclass field simply escaped it. The shadow and its getter are removed; the
annotated base-class property is used instead.

`route.start.exception` holds the actual `Throwable` (`InternalRouteStartupManager` puts it there), so Jackson
was serialising the whole object graph — nested causes, every stack frame with class/file/line, classloader
names and the JDK version.

**No public API change:** `RouteDetailsEndpointInfo.getProperties()` was an override of a method the public
base class `RouteEndpointInfo` also declares, so it stays available on the subclass by inheritance — source
and binary compatible.

### Tests

`CamelRoutesEndpointStartExceptionTest` — 2 cases over the existing `controlled-bar` fixture (a route that
fails to start): neither the info nor the detail view serialises `route.start.exception`, and both still
serialise the benign route properties.

The supervising controller starts routes asynchronously, so a `@BeforeEach` uses Awaitility to wait until the
raw route properties actually hold a `Throwable` under `route.start.exception` before either view is built —
otherwise the assertions could pass with nothing to leak.

Verified meaningful: the detail case fails against the unpatched code and passes with it; the info case passes
either way, which also confirms `@JsonIgnoreProperties` genuinely filters map entries, so removing the shadow
is sufficient. Full `core/camel-spring-boot` suite: 138 tests, 0 failures.

### Changed after review

This PR originally also restricted the health-check `error.stacktrace` detail to the `full` exposure level.
**That has been dropped**, after @luigidemasi pointed out it was not accidental placement:

- CAMEL-18832 created `CamelHealthHelper` explicitly to *"align output to be similar to microprofile-health"*,
  and its original implementation already put the error block outside the exposure-level branch — the `full`
  check governs metadata filtering of the details map, not the error block.
- `camel-microprofile-health`'s `testExposureLevelDefault` asserts `error.stacktrace` **is** present at the
  default level.
- The exposure-level docs say `default` "only include detailed information for health check that was DOWN".

Restricting it only in Spring Boot would silently diverge the two runtimes. The unauthenticated-exposure
concern is real but belongs with CAMEL-24498, where the observability starter raises
`camel.health.exposure-level` to `full` and `show-details` to `always` on an all-interfaces listener — at
Spring Boot's own default (`show-details=never`) these details are not exposed at all.

The `error.message` overwrite across multiple DOWN checks that @luigidemasi also found is a pre-existing
defect independent of this PR, now tracked as
[CAMEL-24512](https://issues.apache.org/jira/browse/CAMEL-24512).

---


davsclaus (approved):
Reviewed against this project's build/test/contribution rules. I checked out the branch locally, built the module, ran the new tests, and — importantly — reverted the two fix files back to `main` while keeping the new tests to confirm they actually catch the described regressions before trusting the PR's own claims about them.

**Verification performed**
- Build: `mvn -DskipTests install` in `core/camel-spring-boot` succeeds.
- New tests: `CamelHealthHelperExposureTest` (3 cases) and `CamelRoutesEndpointStartExceptionTest` (2 cases) all pass on the PR branch.
- Regression check: reverted `CamelHealthHelper.java` and `CamelRoutesEndpoint.java` to `main` while keeping the new tests — both defect tests failed. The failure output is a striking confirmation of the bugs: the health check test leaked a full stack trace at the default exposure level, and the route detail test serialized the entire `route.start.exception` object graph (nested causes, every stack frame with classloader name, module name, JDK version, etc.), exactly as described in the PR body. Restored the fix and confirmed green again.
- Full suite: `mvn test` in `core/camel-spring-boot` — 141 tests, 0 failures, matching the PR description exactly.
- History check: `git show ebccbfe5408` (CAMEL-20993) confirms the `@JsonIgnoreProperties(value = {"route.start.exception"})` annotation and the shadowing `properties` field in `RouteDetailsEndpointInfo` were introduced in the same commit — the filter escaping the subclass looks like an accidental oversight, not an intentional design choice being reverted.
- "No public API change" claim: verified the public base class `RouteEndpointInfo` still declares both the annotated field and `getProperties()`, so removing the subclass's duplicate override is source/binary compatible as claimed.

**Findings**

No blocking issues. This is a clean, well-scoped, well-verified information-disclosure fix.

- Both root causes are independently verified as real defects, not reverts of intentional behavior.
- Test coverage is genuinely meaningful — empirically confirmed the tests fail without the fix.
- Single commit, correctly formatted per project guidelines (`CAMEL-24499: ...`).
- No public API changes, no new dependencies, no Lombok/Records — compliant with project standards.
- Tight scope: two related root causes in the same actuator exposure area, no drift into unrelated code.

No nitpicks or open questions on this one.

luigidemasi:
## Review verdict: Request changes

  Reviewed exact head `2f56679689c5811227e890dd8ee5b3ea173a5a7d` against base `438b015ea3aaff65cca047a318f9385e835bf18e`.

  ### Findings

  #### [P1] Do not silently redefine the established default health-exposure contract

  [`CamelHealthHelper.java:62-82`](https://github.com/apache/camel-spring-boot/blob/2f56679689c5811227e890dd8ee5b3ea173a5a7d/core/camel-spring-boot/src/main/java/org/
  apache/camel/spring/boot/actuate/health/CamelHealthHelper.java#L62-L82)

  CAMEL-18832 introduced this helper to align Spring Boot with MicroProfile health. MicroProfile’s default-exposure regression still explicitly requires
  `error.stacktrace`, while Camel’s documentation says `default` includes detailed information for failed checks.

  This PR removes that longstanding JSON field only from Spring Boot and characterizes its placement as accidental without acknowledging the compatibility change or cross-
  runtime divergence.

  References:

  - [CAMEL-18832 implementation](https://github.com/apache/camel-spring-boot/commit/22a6e0429b9fc995594cf334298c7f617027d1e6)
  - [Current MicroProfile default-exposure test](https://github.com/apache/camel/blob/9b26b811fd2798bda130d441538cd339d8ccf4e2/components/camel-microprofile/camel-
  microprofile-health/src/test/java/org/apache/camel/microprofile/health/CamelMicroProfileHealthCheckTest.java#L324-L363)
  - [Exposure-level documentation](https://github.com/apache/camel/blob/9b26b811fd2798bda130d441538cd339d8ccf4e2/docs/user-manual/modules/ROOT/pages/health-
  check.adoc#L291-L299)

  Either retain the established default behavior or explicitly approve and document the compatibility change, including the resulting MicroProfile divergence.

  #### [P2] Preserve the message for every failing health check

  [`CamelHealthHelper.java:63-77`](https://github.com/apache/camel-spring-boot/blob/2f56679689c5811227e890dd8ee5b3ea173a5a7d/core/camel-spring-boot/src/main/java/org/
  apache/camel/spring/boot/actuate/health/CamelHealthHelper.java#L63-L77)

  Default exposure can return multiple DOWN results. The indicator applies them to one `Health.Builder`, but each result writes to the same top-level `error.message`.
  Spring Boot uses `Map.put`, so the last message replaces every earlier one.

  An exact-head probe with two failures produced:

  ```text
  error.message=second-message
  first.data={error.type=IllegalArgumentException}
  second.data={error.type=IllegalStateException}
  ```

  `first-message` is lost. Before this PR, it remained available inside `first.data.error.stacktrace`.

  Store `error.message` inside each check-scoped data map—retaining the top-level key if compatibility requires it—and add a regression test with two distinct DOWN
  results.

  References:

  - [Indicator loop](https://github.com/apache/camel-spring-boot/blob/2f56679689c5811227e890dd8ee5b3ea173a5a7d/core/camel-spring-boot/src/main/java/org/apache/camel/
  spring/boot/actuate/health/CamelHealthCheckIndicator.java#L52-L69)
  - [Spring Boot detail replacement semantics](https://github.com/spring-projects/spring-boot/blob/v4.1.0/module/spring-boot-health/src/main/java/org/springframework/boot/
  health/contributor/Health.java#L231-L241)

  #### [P2] Establish the asynchronous route-failure precondition in the test

  [`CamelRoutesEndpointStartExceptionTest.java:33-70`](https://github.com/apache/camel-spring-boot/blob/2f56679689c5811227e890dd8ee5b3ea173a5a7d/core/camel-spring-boot/
  src/test/java/org/apache/camel/spring/boot/actuate/endpoint/CamelRoutesEndpointStartExceptionTest.java#L33-L70)

  The supervising controller starts asynchronously after a configured 100 ms delay. The tests immediately serialize the route and only assert that `route.start.exception`
  is absent.

  If serialization occurs before the failed start installs that property, both tests pass against the unfixed implementation without exercising the leak.

  Use Awaitility with an explicit `atMost` timeout to:

  1. Wait until the raw route properties contain a `Throwable` under `route.start.exception`.
  2. Build the endpoint view.
  3. Verify serialization omits that key while preserving benign properties.

  #### [P3] Remove or implement the server-log guarantee

  [`CamelHealthHelper.java:66-76`](https://github.com/apache/camel-spring-boot/blob/2f56679689c5811227e890dd8ee5b3ea173a5a7d/core/camel-spring-boot/src/main/java/org/
  apache/camel/spring/boot/actuate/health/CamelHealthHelper.java#L66-L76)

  The comment states that the full trace remains in the server log, but neither the helper nor its caller logs or attaches the throwable using `down(Throwable)`.

  Spring Boot only logs exceptions attached to its builder or thrown by the indicator. Some Camel health checks merely store exceptions in their result, so the trace is
  not guaranteed to exist in server logs.

  [Spring Boot exception-logging path](https://github.com/spring-projects/spring-boot/blob/v4.1.0/module/spring-boot-health/src/main/java/org/springframework/boot/health/
  contributor/AbstractHealthIndicator.java#L76-L93)

  Remove or qualify the claim unless intentional, non-repetitive logging is implemented.

  ### Verification

  - Exact PR head: both new test classes passed — 5 tests, 0 failures.
  - The same tests applied to the base produced the two intended regression failures.
  - All hosted CI checks currently pass.
  - The route DTO implementation itself behaves correctly under Jackson 2 and Jackson 3.
  - The inherited `getProperties()` remains source- and binary-callable.
  - No code, PR metadata, comments, or reviews were modified during this review.
  
  This review does not replace static analysis or specialized review tooling.

  _Generated by Codex on behalf of Luigi De Masi. This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

oscerd:
Thanks @luigidemasi — this was a good catch on P1, and I checked each point rather than taking them as read. Head is now `85b446273ae`.

**[P1] Default health-exposure contract — agreed, change dropped.**

You're right, and my framing of it as accidental placement was wrong. I verified it: the original CAMEL-18832 implementation (`22a6e0429b9`) already had the error block outside the exposure-level branch — the `full` check governs metadata filtering of the details map, not the error block. The commit message says "Aligned output to be similar to microprofile-health", and `camel-microprofile-health`'s `testExposureLevelDefault` does assert `error.stacktrace` at the default level. The docs agree: `default` "only include detailed information for health check that was DOWN".

I checked git history for the route-detail half (CAMEL-20993) but not for the health half, which is exactly how I ended up reverting an intentional decision. The health change is removed from this PR, along with its test; the PR is now the route-detail fix only.

The underlying exposure concern is real but belongs with CAMEL-24498, where the observability starter raises `camel.health.exposure-level` to `full` and `show-details` to `always` on an all-interfaces listener. At Spring Boot's own default (`show-details=never`) these details are not exposed at all. If the cross-runtime default should change, that needs a coordinated change with camel-microprofile rather than a unilateral one here — happy to open that if you and @davsclaus think it's worth it.

**[P2] `error.message` overwrite — confirmed, filed separately.**

Verified: `CamelHealthCheckIndicator.doHealthCheck` passes the same `Health.Builder` to `applyHealthDetail` for every result, and `error.message` is a flat top-level key, so `Map.put` semantics mean the last DOWN check wins. It pre-exists on main and is independent of this PR, so rather than widen this changeset I've filed [CAMEL-24512](https://issues.apache.org/jira/browse/CAMEL-24512) with your reproduction and the suggested fix (move it into the check-scoped `data` map, regression test with two DOWN results).

**[P2] Asynchronous route-failure precondition — fixed.**

Fair point. The tests did fail against the unfixed implementation when I ran them, but that was timing-dependent rather than guaranteed. A `@BeforeEach` now uses Awaitility (`atMost` 30s) to wait until the raw route properties actually hold a `Throwable` under `route.start.exception` before either view is built, and both tests additionally assert the benign properties are still serialized. Re-verified against the base: the detail test still fails there, so the regression coverage is intact.

**[P3] Server-log guarantee — gone with the health change.**

The comment making that claim was part of the reverted hunk, so it no longer exists. You were right that nothing guaranteed it.

Full `core/camel-spring-boot` suite after the changes: 138 tests, 0 failures. I've left the conversations open for you to resolve.

luigidemasi (approved):
Thanks for addressing the feedback. I rechecked exact head `85b446273ae0aba20f749e5fbf08f6a83f58d23e`.

  The PR-specific findings are resolved:

  - The incompatible health-exposure change and unsupported logging claim were removed.
  - The Awaitility precondition makes the route regression test non-vacuous.
  - The updated test passes on the PR head and fails as expected on the base implementation.
  - The pre-existing multi-DOWN `error.message` issue is appropriately tracked separately in [CAMEL-24512](https://issues.apache.org/jira/browse/CAMEL-24512).

  I have no remaining blocking findings. Thanks for the thorough follow-up.

  _Generated by Codex on behalf of Luigi De Masi. This response was generated by an AI agent and may contain inaccuracies._

davsclaus (approved):

PR #1900, opened 2026-08-26
CAMEL-24508: vault and secrets starters - fail closed when early property resolution fails

oscerd:
### Changed after review

Addressing @davsclaus's findings:

- **`SpringBootCloudConfigPropertiesParser` no longer says "from the vault".** Spring Cloud Config isn't a
  vault — that was a wording defect from applying the same message text uniformly across all seven parsers.
- **`camel.vault.ignore-resolution-failures` is now declared in config metadata.** It's read from the
  `Environment` before any bean exists, so it can't hang off a `@ConfigurationProperties` bean like the
  per-vault options; it's declared in `core/camel-spring-boot`'s `additional-spring-configuration-metadata.json`
  (the mechanism already used there for `management.info.camel.enabled` and
  `management.server.accesslog.enabled`) and regenerates into `spring-boot.json`, so it surfaces in IDE
  completion.

Two of his findings are deliberately not addressed here: replicating the regression test across the other six
parsers (worth doing, kept out to keep this changeset reviewable — happy to do it here or as a follow-up), and
the upgrade-guide entry, which will be opened against `apache/camel` before this merges.

---

### DIFF
--- a/components-starter/camel-aws-secrets-manager-starter/src/main/java/org/apache/camel/component/aws/secretsmanager/springboot/SpringBootAwsSecretsManagerPropertiesParser.java
+++ b/components-starter/camel-aws-secrets-manager-starter/src/main/java/org/apache/camel/component/aws/secretsmanager/springboot/SpringBootAwsSecretsManagerPropertiesParser.java
@@ -44,6 +44,10 @@ public class SpringBootAwsSecretsManagerPropertiesParser implements ApplicationL
     public void onApplicationEvent(ApplicationEnvironmentPreparedEvent event) {
         SecretsManagerClient client;
         ConfigurableEnvironment environment = event.getEnvironment();
+        // an unresolved placeholder would otherwise stay in the property value and become the effective
+        // secret, so resolution failures abort startup unless the operator opts back into the old behaviour
+        final boolean ignoreResolutionFailures
+                = Boolean.parseBoolean(environment.getProperty("camel.vault.ignore-resolution-failures"));
         if (Boolean.parseBoolean(environment.getProperty("camel.component.aws-secrets-manager.early-resolve-properties"))) {
             String accessKey = environment.getProperty("camel.vault.aws.accessKey");
             String secretKey = environment.getProperty("camel.vault.aws.secretKey");
@@ -94,8 +98,17 @@ public class SpringBootAwsSecretsManagerPropertiesParser implements ApplicationL
                                         .replace("}}", ""));
                                 props.put(key, element);
                             } catch (Exception e) {
-                                // Log and do nothing
-                                LOG.debug("failed to parse property {}. This exception is ignored.", key, e);
+                                if (ignoreResolutionFailures) {
+                                    LOG.warn("Failed to resolve property {} from the vault; the placeholder is left "
+                                             + "unresolved because camel.vault.ignore-resolution-failures is enabled", key, e);
+                                } else {
+                                    throw new RuntimeCamelException(
+                                            "Failed to resolve property " + key + " from the vault. Startup is aborted so "
+                                                            + "that the unresolved placeholder cannot become the effective "
+                                                            + "value; set camel.vault.ignore-resolution-failures=true to "
+                                                            + "continue anyway.",
+                                            e);
+                                }
                             }
                         }
                     });
--- a/components-starter/camel-azure-key-vault-starter/src/main/java/org/apache/camel/component/azure/key/vault/springboot/SpringBootAzureKeyVaultPropertiesParser.java
+++ b/components-starter/camel-azure-key-vault-starter/src/main/java/org/apache/camel/component/azure/key/vault/springboot/SpringBootAzureKeyVaultPropertiesParser.java
@@ -44,6 +44,10 @@ public class SpringBootAzureKeyVaultPropertiesParser implements ApplicationListe
     public void onApplicationEvent(ApplicationEnvironmentPreparedEvent event) {
         SecretClient client;
         ConfigurableEnvironment environment = event.getEnvironment();
+        // an unresolved placeholder would otherwise stay in the property value and become the effective
+        // secret, so resolution failures abort startup unless the operator opts back into the old behaviour
+        final boolean ignoreResolutionFailures
+                = Boolean.parseBoolean(environment.getProperty("camel.vault.ignore-resolution-failures"));
         if (Boolean.parseBoolean(environment.getProperty("camel.component.azure-key-vault.early-resolve-properties"))) {
             String vaultName = environment.getProperty("camel.vault.azure.vaultName");
             String clientId = environment.getProperty("camel.vault.azure.clientId");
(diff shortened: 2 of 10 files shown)


davsclaus (approved):
Reviewed against this project's build/test/contribution rules. I checked out the branch, built all 7 affected starter modules, ran the new test, reverted the fix to confirm it's a genuine regression test, restored it, and ran the existing testcontainers-based happy-path test to make sure successful resolution is unaffected.

**Verification performed**
- Build: all 7 modified starters (`camel-aws-secrets-manager-starter`, `camel-azure-key-vault-starter`, `camel-cyberark-vault-starter`, `camel-google-secret-manager-starter`, `camel-hashicorp-vault-starter`, `camel-ibm-secrets-manager-starter`, `camel-spring-cloud-config-starter`) build cleanly with `mvn -DskipTests install`.
- New test: `EarlyResolutionFailureTest` (hashicorp-vault-starter) — 2/2 pass.
- Regression check: reverted `SpringBootHashicorpVaultPropertiesParser.java` to `main` while keeping the new test — the "aborts startup" case failed exactly as expected ("Expected RuntimeCamelException to be thrown, but nothing was thrown"), confirming the test is meaningful. Restored the fix — green again.
- No regression on the happy path: ran the existing testcontainers-based `EarlyResolvedPropertiesTest` for hashicorp-vault (Docker available) — still passes, so a successfully-resolved secret is unaffected by the new fail-closed logic.
- History/precedent: the identical `// Log and do nothing` / `LOG.debug(...)` pattern is confirmed present pre-change in all 7 parsers, so this isn't reverting an intentional design decision — it closes a real silent-failure gap uniformly.

**Findings**

No correctness issues found — the fix behaves as described everywhere I checked it. A couple of non-blocking suggestions:

1. **Test coverage gap across 6 of the 7 modified parsers.** Only `camel-hashicorp-vault-starter` gets a regression test; the other six (`aws-secrets-manager`, `azure-key-vault`, `cyberark-vault`, `google-secret-manager`, `ibm-secrets-manager`, `spring-cloud-config`) are separate, hand-duplicated classes with no shared base, so a mistake in any of them wouldn't be caught by CI. The hashicorp test shows the pattern needs no Docker/network — worth replicating for at least a few more of these, ideally all, given this is a security-relevant, uniformly-applied change.
2. **Upgrade-guide entry outstanding.** This is an intentional breaking change (silent fallback → fail-fast by default). The PR body already flags that the companion `apache/camel` upgrade-guide PR hasn't been opened yet — worth making sure that lands before/at release so it doesn't get lost.
3. **Minor:** the new `camel.vault.ignore-resolution-failures` flag has no `@ConfigurationProperties` field anywhere (unlike every per-vault `camel.vault.*` option), so it won't surface in Spring Boot config metadata/IDE autocomplete. Consider documenting it explicitly.
4. **Minor:** `SpringBootCloudConfigPropertiesParser`'s new messages say "...from the vault" even though Spring Cloud Config isn't a vault — a small wording inconsistency from applying the same text uniformly.

oscerd:
Thanks @davsclaus — I've addressed the two substantive points. Head is now `801da0c94f1` (rebased onto current `main`, which is why it's a force update).

**4. "from the vault" wording in `SpringBootCloudConfigPropertiesParser` — fixed.**

You're right, that was a wording defect I introduced by applying the same text uniformly across all seven parsers. Spring Cloud Config isn't a vault. Its messages now say "from Spring Cloud Config", and the comment above the flag says "effective configuration value" rather than "effective secret".

**3. `camel.vault.ignore-resolution-failures` not in config metadata — fixed.**

Good catch. The flag is read from the `Environment` at `ApplicationEnvironmentPreparedEvent` time, before any bean exists, so it can't hang off a `@ConfigurationProperties` bean the way the per-vault options do. I've declared it in `core/camel-spring-boot`'s `additional-spring-configuration-metadata.json` instead, which is the mechanism already used there for exactly this case (`management.info.camel.enabled`, `management.server.accesslog.enabled`). It regenerates into `spring-boot.json` alongside the other `camel.vault.*` entries, so it now shows up in IDE completion.

**1. Test coverage across the other six parsers — noted, not done here.**

Agreed it's worth doing, and you're right that the pattern needs no Docker. I've left it out of this PR to keep the changeset reviewable rather than adding six near-identical test classes to a security-relevant change that's already touching seven modules. Happy to follow up with a separate PR, or to add them here if you'd prefer them to land together — your call.

**2. Upgrade-guide entry — still outstanding, deliberately.**

The companion `apache/camel` PR for `camel-4x-upgrade-guide-4_23.adoc` isn't open yet. I'd rather write it against the wording that actually lands here than pre-empt review. It will be opened before this merges.

Verification after the changes: `EarlyResolutionFailureTest` 2/2 green; full reactor build 394 modules SUCCESS. The one reactor failure is `camel-toon-starter`, which is unrelated to this PR — it was added to `main` in `978a820dd46` and depends on `org.apache.camel:camel-toon:4.23.0-SNAPSHOT`, which isn't published to apache.snapshots yet; it fails the same way on a clean `main`.

Conversations left open for you to resolve.

oscerd:
Rebased onto `main` (force-update) now that #1905 has landed.

That PR removed six starters whose components were deleted from camel core, which was failing documentation validation on `main` — this branch was inheriting that failure through no fault of its own (`target of xref not found: next@components::ironmq-component.adoc`). With the rebase, `check-doc` should go green here.

No content change: the diff against `main` is byte-for-byte what it was before the rebase, and the module tests were re-run against the new `main` before pushing.

Croway:
@oscerd I think there is room for improvement here, in particular wrt code duplication, I think we might create an abstract `EarlyResolutionPropertiesParser` (or something similar, naming things is hard) in `core/camel-spring-boot` and refactor the PropertiesParser, can be done in another PR.

Moreover I have some agents findings (I didn't verify them though)
```
- Precedence is inverted. getPropertySources() iterates highest-precedence first, and all seven put into one flat Properties. Last write wins, so the lowest-precedence source's value survives, and then the merged result is addFirsted above everything. If the same key holds a vault placeholder in both application.properties and an application-prod.properties, the wrong one wins. Fix is to iterate in reverse.
- replace() should be substring(). .replace("{{aws:", "").replace("}}", "") strips every occurrence, not just the delimiters. A remainder containing }} gets silently mangled.
- First failure aborts. Introduced by this PR. An operator with five broken placeholders restarts five times to find them all. Collecting the failures and throwing one aggregate is barely more code and much kinder.
- Boolean.parseBoolean(getProperty(...)) vs getProperty(key, Boolean.class, false). With parseBoolean, ignore-resolution-failures=yes silently means false. That fails closed, so it's the safe direction, but for a flag whose whole purpose is opting out of a security behaviour I'd rather a typo produced a conversion error than silent strictness. Your call, worth being deliberate about it.
- The flag is read before the early-resolve guard in all seven, so it's evaluated even when the feature is off. Harmless, reads oddly, disappears under the abstraction.
- "Fail closed" is narrower than the PR body implies. The match is startsWith(prefix) && endsWith("}}"), so only values that are entirely a placeholder are handled. url=https://host/{{aws:path}}/x is never touched here and falls through to normal Camel resolution. Worth stating precisely in the upgrade guide, since that's what people will read to understand the breaking change.
- Google's parser wraps IOException in plain RuntimeException, not RuntimeCamelException, unlike its own sibling branch two lines down.
- Worth checking: PropertiesFunction has an optional(remainder) default method for "missing is acceptable." The early resolver ignores it, so a placeholder Camel would treat as optional at normal resolution time now aborts startup at early resolution time. I haven't traced whether any vault function overrides it, but the semantics should probably line up.
```

oscerd:
I would open a Jira issue with this improvement and findings. Thanks @croway

Croway:
will do it

Croway (approved):

PR #1902, opened 2026-08-26
Removal of component iec60870

apupier:
removed in camel core https://github.com/apache/camel/pull/25757

PR #1905, opened 2026-08-27
Removal of components deleted from camel core: IronMQ, DigitalOcean, IRC, json-patch, Splunk, langchain4j-tools

oscerd:
Six components were removed from camel core; their starters and starter docs are still here, and those docs
carry `xref:{csb-camel-components}::<name>-component.adoc` targets that no longer exist. **Documentation
validation is currently failing on `main` because of this.**

| starter removed | removed in camel core |
|---|---|
| `camel-ironmq-starter` | [apache/camel#25769](https://github.com/apache/camel/pull/25769) — deprecated in 4.21, client library unmaintained since 2017 |
| `camel-digitalocean-starter` | `4fccebb8da41` |
| `camel-irc-starter` | `933913a0ce20` |
| `camel-json-patch-starter` | `2c538511af7f` |
| `camel-splunk-starter` | `53912a500d66` |
| `camel-langchain4j-tools-starter` | `2a2b0e1bb71c` |

**Deliberately kept**, because those components still exist upstream:

- `camel-splunk-hec-starter` — only `camel-splunk` was removed, not `camel-splunk-hec`
- `camel-langchain4j-{agent,chat,embeddings,embeddingstore,tokenizer,web-search}-starter` — only
  `camel-langchain4j-tools` was removed

Same file set per component as the recent removals (#1899 LevelDB, #1902 iec60870): catalog entry and JSON,
the starter module, the `components-starter` module list, the nav entry and starter page, the BOM and
dependencies entries, and the integration test.

### Why these are batched rather than one PR each

They started as one PR per component, matching #1899/#1902. That does not work here: the published `next`
docs catch up component by component, so each single-component PR goes green on its own component and then
immediately fails on the next one. This PR went through exactly that cycle twice before I checked the whole
set at once.

### Verification

Rather than reacting to whatever CI surfaces next, every component xref under `docs/` was cross-checked
against the 411 component pages that actually exist on camel `main`:

```
checked 377 xref targets, missing 0
```

Also: `mvn validate` on the full reactor passes, `camel-catalog-provider-springboot` builds, all poms parse,
and the BOM and dependencies files lose exactly 5 lines per component and nothing else.

### Note on the build job

The `build` job failure is the pre-existing `Cannot run program "./mvnw"` breakage in `camel-itest-spring-boot`,
which hits any PR touching `tests/` — every component removal deletes an IT, so it hits all of them. It is
identical on #1902, which was merged with that same failure. Not introduced here.

---


Croway (approved):

davsclaus (approved):

apupier (approved):

oscerd:
**`check-doc` is now green** — which was the point of this PR. Documentation validation passes with all six starters removed.

It took three iterations to get there, and the first two are worth recording because the failure mode is misleading:

1. **ironmq only** — that xref cleared, then `digitalocean`, `irc`, `json-patch` and `splunk` appeared.
2. **+ those four** — they cleared, then `langchain4j-tools` appeared.

The published `next` docs catch up one component at a time, so a green run only proves nothing has propagated *yet*. Rather than keep reacting, every component xref under `docs/` was cross-checked against the 411 component pages that actually exist on camel `main`:

```
checked 377 xref targets, missing 0
```

That is the basis for believing this set is complete, rather than just currently quiet.

**The `build` job failure is not from this PR.** It is 630 occurrences of `Cannot run program "./mvnw"` in `camel-itest-spring-boot` archetype tests — no compilation errors, no dependency resolution failures. It hits any PR touching `tests/`, and every component removal deletes an integration test. #1902 failed identically and was merged.

Two things deliberately **not** removed, since those components are still present in camel core:

- `camel-splunk-hec-starter` — only `camel-splunk` was removed upstream
- the six sibling `camel-langchain4j-*-starter` modules — only `camel-langchain4j-tools` was removed

PR #1912, opened 2026-08-28
CAMEL-24504: camel-spring-boot - extend the String conversion guard beyond InputStream

oscerd:
CAMEL-23378 (`52b017db006`) added this guard to `SpringTypeConverter.convertTo()`:

```java
// do not attempt to convert String -> InputStream (or subclasses like FileInputStream).
// Spring's ObjectToObjectConverter finds FileInputStream(String) constructor and treats
// the String value as a file path instead of data content.
if (value instanceof String && InputStream.class.isAssignableFrom(type)) {
    return null;
}
```

The reasoning generalises: Spring's `ObjectToObjectConverter` will find *any* single-String constructor, and
several of them take a path. Only `InputStream` was covered.

### What actually happens without the guard

Measured by running the new tests against `main`:

| target | behaviour on `main` |
|---|---|
| `FileWriter` | **conversion succeeds** — `expected <null> but was <java.io.FileWriter@...>`. `FileWriter(String)` **creates the named file on disk.** |
| `FileReader` | throws `ConversionFailedException` — Camel's own converters never get a chance |
| `ZipFile` | throws `ConversionFailedException` — same |

The `FileWriter` case is the notable one: a String body is not merely misread as a path, it causes a file to be
created (and truncated, if it already exists).

### Change

`FileReader`, `Writer` and `ZipFile` join `InputStream` in the guard, extracted into a small
`isFileBackedTarget(Class<?>)` helper. The reader check is deliberately limited to `FileReader` rather than all
`Reader` implementations, preserving the safe `String` → `StringReader` conversion where the value is content.

**`java.io.File` is deliberately not included.** `String` → `File` is a documented Spring conversion where the
String genuinely *is* a path, and blocking it would break legitimate use. There is a test pinning that it still
works, so the guard cannot quietly grow to swallow it.

The scan this came from suggested inverting to an allowlist of safe target types. I did not do that — it would
be a much larger behaviour change across every `String` conversion this converter handles, and it is not
needed to close the hazard. Worth considering separately if the denylist keeps growing.

### Tests

Five added to `SpringTypeConverterTest`, following the shape of the two CAMEL-23378 added:

- `FileReader`, `FileWriter` and `ZipFile` targets are blocked
- `String` → `StringReader` remains available
- the `FileWriter` case additionally asserts no file appears on disk
- `String` → `File` still converts

Verified meaningful: the three hazard tests fail against `main` (1 failure, 2 errors); the `File` test passes
either way as a regression guard. Full `core/camel-spring-boot` suite: 143 tests, 0 failures. Focused
`SpringTypeConverterTest` after preserving `StringReader`: 8 tests, 0 failures.


davsclaus (commented):
Reviewed against this repo's rule files (`.oss-ai-helper-rules/`). Solid, well-scoped fix — it generalizes the existing CAMEL-23378 guard in `SpringTypeConverter.convertTo()` to cover `Reader`, `Writer`, and `ZipFile` alongside `InputStream`, all of which Spring's `ObjectToObjectConverter` can misinterpret a String as a file path for. The `FileWriter` case is the standout: without this guard it silently creates a file on disk from a String body. `File` is deliberately left out of the guard since `String -> File` is a legitimate, documented conversion — good call, and it's pinned by a regression test so the guard can't quietly grow to swallow it.

Verified locally: checked out the branch and ran `mvn verify` for `core/camel-spring-boot` — build succeeds, 143 tests / 0 failures, matching the PR description. Also confirmed via `git log`/`git blame` that this extends rather than conflicts with the CAMEL-23378 guard it builds on.

Found one minor import-ordering issue in the new test imports and pushed a fix commit directly to this branch (with maintainer-edit enabled and by agreement with the branch owner) rather than leaving it as a suggestion, since it was purely mechanical:

- `SpringTypeConverterTest.java`: the new imports (`File`, `FileReader`, `FileWriter`, `Reader`, `Files`, `Path`, `ZipFile`) were inserted above the pre-existing `import java.io.InputStream;`, leaving it out of alphabetical order after `java.util.zip.ZipFile`. Re-sorted alphabetically; reran `SpringTypeConverterTest` after the fix — 8/8 tests still pass.

No other conventions, coverage, or scope issues found. This review does not replace CodeRabbit/Sourcery/SonarCloud or a full static analysis pass.

Croway (approved):

Croway (approved):

PR #1927, opened 2026-09-02
CAMEL-24498: align camel-observability-services-starter defaults with the Spring Boot baseline

Croway:
Fixes [CAMEL-24498](https://issues.apache.org/jira/browse/CAMEL-24498).

## What

`ObservabilityServicesEnvironmentPostProcessor` injects a set of management defaults as soon as `camel-observability-services-starter` is on the classpath. Three of them were wider than the Spring Boot or Camel setting they replaced, and this PR brings them back in line.

| Property | Before | After |
| --- | --- | --- |
| `management.server.address` | not set (listener on all interfaces) | `127.0.0.1` |
| `management.endpoint.health.show-details` | `always` | `when-authorized` |
| `camel.health.exposure-level` | `full` | not injected (Camel default `default`) |

Everything else the starter injects is unchanged.

## Why

- **`management.server.port` was injected with no matching `management.server.address`.** Spring Boot ships no separate management listener at all, so the starter both opens the listener *and*, implicitly, put it on every interface. Opening it is the starter's job; deciding how far it reaches should be the operator's. `camel-jolokia-starter` binds its agent to loopback for the same reason.
- **`show-details=always`** is two steps past the Spring Boot default of `never`. Camel health checks report on the resources a route talks to — broker connections, data sources, remote endpoints — and their detail can identify those resources. `when-authorized` shows the indicators to an authenticated caller and the overall status alone to everybody else; with no Spring Security on the classpath it behaves as `never`.
- **`camel.health.exposure-level=full`** overrode the Camel default of `default`. At `default`, `CamelHealthHelper` filters health check *metadata* (endpoint URIs, route and consumer identifiers) out of the response and keeps check names, error messages and stack traces — so failure diagnosis is unaffected. `full` is now documented as an opt-in.

### Why the `live`/`ready` groups keep `show-details=always`

The JIRA left this conditional on whether the Kubernetes probe flow needs it. It does, and it costs nothing:

- The kubelet reads `/observe/health/live` and `/observe/health/ready` **unauthenticated**, so `when-authorized` would collapse to `never` for them.
- The kubelet puts the (truncated) HTTP response body into the probe-failure event, so `kubectl describe pod` names the indicator that took the pod down. With `never` the operator only sees the status code.
- Both groups contain availability-state indicators only. `CamelLivenessStateHealthIndicator` / `CamelReadinessStateHealthIndicator` extend Spring's `LivenessStateHealthIndicator` / `ReadinessStateHealthIndicator`, which report a status and contribute no details. So `always` on those groups reveals the indicator names and their UP/DOWN state, and nothing else.

## Behaviour change and how to opt back in

The property source is still added with `addLast`, so every one of these is overridden by ordinary application configuration.

**Kubernetes deployments whose kubelet probes or Prometheus scrapers reach the pod over the network must now widen the bind address explicitly:**

```properties
management.server.address = 0.0.0.0
```

The other two:

```properties
management.endpoint.health.show-details = always
camel.health.exposure-level = full
```

This is called out in the starter docs and needs a note in the 4.23 upgrade guide in `apache/camel` (drafted separately, not part of this PR).

## Docs

The starter had no hand-written doc sections. This PR adds `src/main/doc/intro.adoc` and `src/main/doc/usage.adoc`, which the generator folds into `docs/spring-boot/modules/ROOT/pages/starters/observability-services.adoc` (regenerated and committed). They document the endpoint layout, **the full injected property set as a table**, and a section each on the bind address, health detail exposure and the Camel exposure level.

A test asserts that the injected key set matches the documented table, so the two cannot drift apart silently.

## Tests

`mvn install -pl components-starter/camel-observability-services-starter` — 10 tests, 0 failures.

- `ObservabilityServicesEnvironmentPostProcessorTest` (7) — extended with `managementListenerBindsToLoopback`, `aggregateHealthDetailsRequireAuthorization`, `probeGroupsKeepTheirDetails`, `camelHealthExposureLevelIsNotForced` and `injectedPropertiesAreTheDocumentedSet`; the existing assertions for the removed/changed values were updated.
- `ObservabilityServicesOptInOverridesTest` (3, new) — verifies each of the three settings can be opted back into from application configuration.

No `Thread.sleep` introduced; the tests are assertions over the `Environment`, so no waiting is involved.


oscerd (approved):

oscerd:
Is `management.server.address` now defaulted to loopback for the other components as well, or only here?

As far as I can see this starter is the only thing in the repo that opens a separate management listener, so there would be nothing to default elsewhere: `management.server.port` and `management.server.address` are set only in `ObservabilityServicesEnvironmentPostProcessor`, the other two `EnvironmentPostProcessor` implementations (`CamelCloudConfigEnvironmentPostProcessor`, `CamelVirtualThreadEnvironmentPostProcessor`) touch no management properties, and nothing references `management.port` or `ManagementServerProperties`. The other `management.server.*` hits are `management.server.accesslog.*`, which is our own access-log config rather than the bind address.

Spring Boot's own default is no separate management port at all, with actuator served on the main application port that the app already binds deliberately, and `management.server.address` only takes effect once a separate port is configured. So unless I have missed a starter that opens one, this is the only place that needed it.

Worth confirming though, since if another starter does grow a management listener later it will inherit the all-interfaces default again.

_Comment by Claude Code on behalf of Andrea Cosentino._

oscerd:
@Croway @squakez

squakez:
My concern is about a couple of additional points.

Do we need to extend this behavior to other components that are exposing an endpoint? The `camel-observability-services` is just a layer on top of existing components, `camel-metrics`, `camel-health`, ... etc.

Do we need to extend this to other runtimes to "harmonize" the security measures in place? The goal of `camel-observability-services` is to provide a sensible default configuration and avoid any external consumer to use it without knowing if the runtime it's coming is Spring Boot, Quarkus or Camel Main. It is worthy to clarify this aspect as well.

oscerd:
Good questions, and the second one has a concrete answer that I think turns into work.

**Other components exposing an endpoint**

I would draw the line at "who chose the bind address" rather than "does it listen". `camel-metrics` and `camel-health` do not open a socket themselves; they are SPIs and the runtime's management layer serves them. Components that do open one, like platform-http, undertow, jetty or netty-http, take the host from the endpoint the route author wrote, or from the application's own `server.address`. The user picked that.

What made this starter different is that it injected a *second* listener nobody asked for and chose `0.0.0.0` on the user's behalf. By that test it was the only case in camel-spring-boot: `management.server.port` and `management.server.address` appear nowhere else, and the other two `EnvironmentPostProcessor` implementations touch no management properties. So I would not extend it to the endpoint components.

**Other runtimes**

Here you are right, and both siblings still have the old behaviour.

`camel-quarkus`, `extensions/observability-services/runtime/src/main/resources/application.properties` on current main:

```properties
quarkus.management.enabled=true
quarkus.management.port=9876
```

There is no `quarkus.management.host`, and Quarkus defaults it to `0.0.0.0`. That is the same shape as what we just fixed: an injected management listener on 9876, reachable from anywhere, that the user never asked to expose.

`camel-main`, `HttpManagementServerConfigurationProperties` on current camel main:

```java
@Metadata(defaultValue = "0.0.0.0")
private String host = "0.0.0.0";
@Metadata(defaultValue = "8080")
private int port = 8080;
```

serving `/observe/info`, `/observe/health` and `/observe/jolokia`. That one is arguably wider than the Spring Boot case, because the same server can also carry the dev console, upload, download and send, several of which already carry `@Metadata(security = "insecure:dev")` — so we already treat them as sensitive, we just do not bind them anywhere safe by default.

So after this PR the three runtimes disagree: Spring Boot binds loopback, Quarkus and Main bind all interfaces, for the same `/observe/*` surface. Given that harmonised defaults are the whole point of `camel-observability-services`, that seems worth fixing rather than leaving.

Happy to open the two follow-up issues if you and @Croway agree that is the right direction, or to leave them to whoever owns those runtimes.

_Comment by Claude Code on behalf of Andrea Cosentino._

PR #1931, opened 2026-09-02
CAMEL-24506: camel-jasypt-starter - modern default algorithm and safer usage example

Croway:
## What this fixes

[CAMEL-24506](https://issues.apache.org/jira/browse/CAMEL-24506) — two items in `camel-jasypt-starter`.

### 1. Default algorithm

`JasyptEncryptedPropertiesConfiguration` defaulted `camel.component.jasypt.algorithm` to `PBEWithMD5AndDES` — a 56-bit DES cipher with MD5-based key derivation — and that value flowed into `EnvironmentStringPBEConfig.setAlgorithm` on every decryption path.

The default is now `PBEWITHHMACSHA256ANDAES_256`. No new machinery was needed: the starter already lists that algorithm in `JasyptEncryptedPropertiesUtils.ALGORITHMS_THAT_REQUIRE_IV`, so `getIVGenerator` installs `RandomIvGenerator` automatically (using `camel.component.jasypt.random-iv-generator-algorithm`, default `SHA1PRNG`) whenever `iv-generator-class-name` is not set explicitly.

### 2. Usage example placed the master password beside the ciphertext

`src/main/doc/usage.adoc` showed `camel.component.jasypt.password = the-master-password` in the same properties block as the `ENC(...)` value, which removes the benefit of encrypting it. The example now uses `sysenv:JASYPT_PASSWORD`, carries a `WARNING` admonition that the master password must come from the environment or an external secret store and must never live in the same file or repository as the values it protects, and documents the `sysenv:` / `sys:` prefixes that `JasyptEncryptedPropertiesUtils.parsePassword` already supports.

While in that file, the encryption command was also corrected: `camel jasypt encrypt --password= --input=` is not a real command. It is replaced by the documented `jbang org.apache.camel:camel-jasypt:<camel-version> -c encrypt ...` form, with the algorithm and IV generator flags, plus a note that the CLI entrypoint is deprecated in favour of the Jasypt distribution's `encrypt.sh`.

## Behaviour change and how to opt back

This is a breaking change for values encrypted under the old default. Such a value cannot be decrypted with `PBEWITHHMACSHA256ANDAES_256`; resolving the property throws `EncryptionOperationNotPossibleException`. Either re-encrypt the values, or pin the previous algorithm:

```properties
camel.component.jasypt.algorithm = PBEWithMD5AndDES
```

An upgrade-guide entry has been drafted for the `camel-spring-boot` section of `camel-4x-upgrade-guide-4_23.adoc` in `apache/camel`, and will be submitted separately.

One practical detail that the docs now call out: the Jasypt CLI installs **no** IV generator unless asked (`StandardPBEByteEncryptor` falls back to `NoIvGenerator`), so encryption must be given both the algorithm and a random IV generator, otherwise the value it produces cannot be decrypted at runtime:

```bash
jbang org.apache.camel:camel-jasypt:<camel-version> \
  -c encrypt -p "$JASYPT_PASSWORD" -i my-secret-value \
  -a PBEWITHHMACSHA256ANDAES_256 -riga SHA1PRNG
```

## Upstream coordination

The upstream `camel-jasypt` component does **not** currently match this default. `JasyptPropertiesParser.initEncryptor` only calls `setAlgorithm` when the `algorithm` field is non-null, and it is null by default, so the parser falls back to the Jasypt library default `StandardPBEByteEncryptor.DEFAULT_ALGORITHM` = `PBEWithMD5AndDES`. The same is true of the `camel-jasypt` CLI (`org.apache.camel.component.jasypt.Main`), where `-a` is optional.

The upstream default should follow in a separate change in `apache/camel`; this PR does not make it. Until then the starter and the component diverge, which is why the starter documentation now states explicitly that the encryption tooling must be given a matching `--algorithm`/`-a` (and IV generator) rather than relying on its default.

## Tests

New `JasyptDefaultAlgorithmTest` (`ApplicationContextRunner` over `JasyptEncryptedPropertiesAutoconfiguration`, so it exercises the real bean wiring rather than the configuration object alone):

- `defaultAlgorithmRequiresAndGetsAnInitializationVector` — the configuration and the `EnvironmentStringPBEConfig` bean both carry `PBEWITHHMACSHA256ANDAES_256`, and the IV generator is a `RandomIvGenerator`.
- `defaultAlgorithmEncryptsAndDecrypts` — encrypt/decrypt round trip through the auto-configured `StringEncryptor` under the new default.
- `legacyAlgorithmCanBePinnedExplicitly` — with `camel.component.jasypt.algorithm=PBEWithMD5AndDES`, the IV generator falls back to `NoIvGenerator` and a fixed ciphertext produced under the legacy algorithm decrypts correctly (the opt-back).
- `legacyValueIsNotReadableUnderTheDefaultAlgorithm` — the same ciphertext fails under the new default, pinning down the breaking change the upgrade note describes.

`EncryptedPropertiesUtilsTest.noIvGeneratorPropertyTest` is updated: a default configuration now auto-detects `RandomIvGenerator`. A new `noIvGeneratorPropertyWithAlgorithmThatDoesNotNeedIvTest` keeps the `NoIvGenerator` path covered for an algorithm that does not require an IV.

Result of `mvn install -pl components-starter/camel-jasypt-starter`:

```
Tests run: 71, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Regenerated and committed: `components-starter/camel-jasypt-starter/src/main/docs/jasypt.json` and `docs/spring-boot/modules/ROOT/pages/starters/jasypt.adoc` (the latter is produced from `src/main/doc/*.adoc` by `camel-spring-boot-generator-maven-plugin:update-starter-doc-page`).

### DIFF
--- a/components-starter/camel-jasypt-starter/src/main/doc/usage.adoc
+++ b/components-starter/camel-jasypt-starter/src/main/doc/usage.adoc
@@ -1,14 +1,47 @@
-Add encrypted values to your `application.properties` using the `ENC(...)` syntax:
+Add encrypted values to your `application.properties` using the `ENC(...)` syntax, and point the starter at the master
+password:
 
 [source,properties]
 ----
 my.secret = ENC(encrypted-value-here)
-camel.component.jasypt.password = the-master-password
+camel.component.jasypt.password = sysenv:JASYPT_PASSWORD
 ----
 
-Use Camel's Jasypt tooling to encrypt values:
+WARNING: The master password must never be stored in the same file, or the same repository, as the encrypted values it
+protects. A file that carries both the ciphertext and the key that unlocks it is no better than a plaintext file. Supply
+the master password from the environment or from an external secret store at deployment time.
+
+`camel.component.jasypt.password` understands two prefixes that keep the password out of the configuration file:
+
+* `sysenv:<name>` looks the password up in the OS environment variable `<name>`.
+* `sys:<name>` looks the password up in the JVM system property `<name>`.
+
+Any other value is used as the password verbatim, which is only appropriate when the property itself is injected by an
+external secret store (for example a mounted secret or a config server) rather than checked in.
+
+=== Encryption algorithm
+
+The default algorithm is `PBEWITHHMACSHA256ANDAES_256`. It requires an initialization vector, which the starter
+generates automatically (`org.jasypt.iv.RandomIvGenerator`) unless `camel.component.jasypt.iv-generator-class-name` is
+set explicitly.
+
+Use Jasypt tooling to encrypt values, passing the *same* algorithm and a random IV generator; a value encrypted under a
+different algorithm, or without an IV generator, cannot be decrypted at runtime:
 
 [source,bash]
 ----
-camel jasypt encrypt --password=the-master-password --input=my-secret-value
+jbang org.apache.camel:camel-jasypt:<camel-version> \
+  -c encrypt -p "$JASYPT_PASSWORD" -i my-secret-value \
+  -a PBEWITHHMACSHA256ANDAES_256 -riga SHA1PRNG
+----
+
+NOTE: The `camel-jasypt` CLI entrypoint is deprecated. The `encrypt.sh` script shipped in the
+https://github.com/jasypt/jasypt/releases/tag/jasypt-1.9.3[Jasypt distribution] provides the same workflow, and takes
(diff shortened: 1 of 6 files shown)


oscerd (approved):

Croway:
Camel SNAPSHOT release is failing, therefore the failing CI. locally the test are successful, merging

PR #1934, opened 2026-09-02
CAMEL-24593: platform-http-starter - delete multipart uploads when the exchange is done

Croway:
Fixes https://issues.apache.org/jira/browse/CAMEL-24593

## What

`SpringBootPlatformHttpBinding.populateAttachments()` copies every accepted multipart upload into the servlet
temp directory (`ServletContext.TEMPDIR`) under a random UUID name, and uses that copy as the attachment
`DataSource` and, for a single upload, as the `Path` message body and the `CamelFilePath` header.

`MultipartFile.transferTo()` moves the container's part file, so the container's own end-of-request cleanup no
longer finds anything to delete, and nothing in the starter deleted the copy either. Every upload the
application accepted stayed on disk for the life of the process, including uploads a route only inspected and
discarded.

The copy itself is deliberate - it was introduced in CAMEL-21461 so the body can be a `Path` and the attachment
stays readable after the servlet request completed - but owning the file means removing it.

The other HTTP bindings do not leak: `camel-http-common`'s `DefaultHttpBinding` reads the part through the
container managed file, which the container deletes at the end of the request, and `camel-platform-http-vertx`
has `deleteUploadedFilesOnEnd`, defaulting to `true`.

## How

* The binding collects the temp files it created for a request and, unless the cleanup is turned off, registers
  a `SynchronizationAdapter` through `exchange.getExchangeExtension().addOnCompletion(...)` that deletes them
  when the exchange is done being routed. The attachment `DataSource` and the `Path` body point at the files
  until then, so they cannot be deleted any earlier; the consumer writes the HTTP response before `doneUoW`, so
  the response is already out when the deletion happens.
* Opt-out mirroring the Vert.x implementation:
  `camel.component.platform-http.server.delete-uploaded-files-on-end`, default `true`. The endpoint options
  live in upstream `camel-platform-http` and cannot be extended from this repository, so the option is a new
  starter owned `@ConfigurationProperties` class, `SpringBootPlatformHttpServerProperties`, alongside the
  existing `camel.component.platform-http.server.undertow.accesslog.*`. It is wired through
  `SpringBootPlatformHttpAutoConfiguration` -> `SpringBootPlatformHttpEngine` -> `SpringBootPlatformHttpConsumer`
  onto the binding. Existing public constructors are unchanged; the engine gained an extra constructor that
  defaults to `true`.

## Behaviour change and how to opt out

Uploaded temp files are now removed after the exchange completes. Routes that consume the upload while routing
(file producer, streaming it out, unmarshalling it) are unaffected - routing a `Path` body to a file producer
copies the content. A route that stores the temp path and reads the file *after* the exchange has finished must
set:

```properties
camel.component.platform-http.server.delete-uploaded-files-on-end=false
```

and is then responsible for deleting the file itself. An upgrade guide entry has been drafted for the
`apache/camel` 4.23 upgrade guide and will be submitted separately.

## Tests

* `SpringBootPlatformHttpUploadCleanupTest` - single upload deleted after the exchange is done (also asserting
  the route saw the file, and that `CamelFilePath` and the `Path` body pointed at it), multi attachment upload
  where both copies are deleted, and a multipart request without a file part.
* `SpringBootPlatformHttpUploadCleanupDisabledTest` - with the opt-out the file is still there after the
  response.
* Awaitility is used for the post-response assertions, no `Thread.sleep`.
* Full `camel-platform-http-starter` surefire suite run locally, green (rebased on current `main`).

Docs: `src/main/docs/platform-http.adoc` gained a "File uploads" section; the generated
`src/main/docs/platform-http.json` and `docs/spring-boot/modules/ROOT/pages/starters/platform-http.adoc` were
regenerated by the module build.

### DIFF
--- a/components-starter/camel-platform-http-starter/src/main/docs/platform-http.adoc
+++ b/components-starter/camel-platform-http-starter/src/main/docs/platform-http.adoc
@@ -17,6 +17,27 @@ carry no matrix parameters. A request to `/greeting/John%20Doe;v=1` sets the `na
 The `CamelHttpPath` header is not affected: it reports the raw request path, with the servlet context-path
 removed.
 
+== File uploads
+
+Every accepted multipart file upload is copied out of the servlet container into the servlet temporary directory
+(`jakarta.servlet.ServletContext#TEMPDIR`), so that it can still be read after the HTTP request has completed. The
+copy is what the route sees: it is the `jakarta.activation.DataSource` of the attachment and, when the request carries
+a single file, also the `java.nio.file.Path` message body and the value of the `CamelFilePath` header.
+
+Because Camel owns that copy, it is deleted again once the exchange is done being routed, that is after the response
+has been written. A route that consumes the upload during routing - saving it with the file producer, streaming it to
+a remote system, unmarshalling it - is unaffected, since the content has already been read or copied by then.
+
+Set the following property if the application hands the temporary file over to something that reads it *after* the
+exchange has completed, in which case the application becomes responsible for deleting the file:
+
+[source,properties]
+----
+camel.component.platform-http.server.delete-uploaded-files-on-end=false
+----
+
+This mirrors the `deleteUploadedFilesOnEnd` option of the Vert.x platform-http implementation and defaults to `true`.
+
 == Undertow Access Log
 
 You can enable Undertow access log to be managed by whatever logging library you have in your camel application, you have to set the following parameters:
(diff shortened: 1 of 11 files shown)


oscerd (approved):

Croway:
The CI failure here is `SpringBootPlatformHttpCookiesTest.echoCookie`, which fails on `main` against the current camel-platform-http snapshot since CAMEL-24453 made the request-header echo suppression case-insensitive. The test fix is in #1938; once that is merged the checks here can be re-run without a rebase.

PR #1936, opened 2026-09-02
CAMEL-24512: camel-spring-boot - keep each failing health check's error message

oscerd:
`CamelHealthCheckIndicator.doHealthCheck` loops over every health check result and passes the same
`Health.Builder` to `CamelHealthHelper.applyHealthDetail`. The error message was written only as a flat
top-level detail:

```java
builder.withDetail("error.message", error.getMessage());
```

`Health.Builder.withDetail` is a `Map.put`, so that key is a single slot shared by every check. With two DOWN
checks the actuator response carried only the second message:

```
error.message=second-message
first.data={...}
second.data={...}
```

The per-check `<id>.data` maps stay separate, so the fix is to put the message there too. The top-level key is
kept as it was, for compatibility.

Found by @luigidemasi while reviewing CAMEL-24499; it pre-dates that change and is independent of it.

### Scope

Deliberately confined to `error.message`. #1929 (CAMEL-24592) changes the stack-trace gating in the same
method, and @Croway scoped that PR to leave this alone — doing the same in reverse here, so the two should
apply cleanly in either order.

### Tests

`CamelHealthHelperMultipleFailuresTest`:

- `eachFailingCheckKeepsItsOwnMessage` — two DOWN results with distinct messages, asserting each is retrievable
  under its own check id
- `topLevelErrorMessageIsStillReported` — the compatibility guard on the flat key

Verified meaningful: the first fails against `main` (`expected <first-message> but was <null>`), the second
passes either way. Full `core/camel-spring-boot` suite: 163 tests, 0 failures.

### Note on the root build

The full reactor currently fails on `camel-hivemq-starter`, unrelated to this change:
`org.apache.camel:camel-hivemq:4.23.0-SNAPSHOT` is not on apache.snapshots yet (main's HEAD is the commit that
added the starter). 171 modules build, `core/camel-spring-boot` among them, before the reactor stops there.
Same publishing-lag shape as `camel-toon` a few days ago.

https://claude.ai/code/session_0165HC1XCB3h6mMii6wbZneG

### DIFF
--- a/core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/actuate/health/CamelHealthHelper.java
+++ b/core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/actuate/health/CamelHealthHelper.java
@@ -61,7 +61,11 @@ final class CamelHealthHelper {
 
             result.getError().ifPresent(error -> {
                 if (error.getMessage() != null) {
+                    // the top-level key is kept for compatibility, but it is a single slot on a builder shared
+                    // by every check, so with more than one DOWN check the last one wins; the check-scoped copy
+                    // is the one that stays addressable per check
                     builder.withDetail("error.message", error.getMessage());
+                    data.put("error.message", error.getMessage());
                 }
                 final StringWriter stackTraceWriter = new StringWriter();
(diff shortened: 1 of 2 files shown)


Croway (approved):

luigidemasi (approved):
LGTM!

PR #1943, opened 2026-09-07
CAMEL-24293: camel-zipfile-starter - Align ZipFileDataFormatTest

bsmolicek:
Use the full ZIP entry name supplied by CAMEL-24293 when extracting the test fixture. The CamelFileName header is intentionally sanitized to the basename.

### DIFF
--- a/components-starter/camel-zipfile-starter/src/test/java/org/apache/camel/dataformat/zipfile/springboot/ZipFileDataFormatTest.java
+++ b/components-starter/camel-zipfile-starter/src/test/java/org/apache/camel/dataformat/zipfile/springboot/ZipFileDataFormatTest.java
@@ -354,8 +354,7 @@ public class ZipFileDataFormatTest {
                                 @Override
                                 public void process(Exchange exchange) throws Exception {
                                     ZipFile zfile = new ZipFile(new File("src/test/resources/hello.odt"));
-                                    ZipEntry entry = new ZipEntry(
-                                            (String) exchange.getIn().getHeader(Exchange.FILE_NAME));
+                                    ZipEntry entry = new ZipEntry((String) exchange.getIn().getHeader("zipFileName"));
                                     String outputDirectory = "hello_out";
                                     File file = new File(outputDirectory, entry.getName());
 


components-starter/camel-zipfile-starter/src/test/java/org/apache/camel/dataformat/zipfile/springboot/ZipFileDataFormatTest.java
```diff
@@ -354,8 +354,7 @@ public void configure() {
                                 @Override
                                 public void process(Exchange exchange) throws Exception {
                                     ZipFile zfile = new ZipFile(new File("src/test/resources/hello.odt"));
-                                    ZipEntry entry = new ZipEntry(
-                                            (String) exchange.getIn().getHeader(Exchange.FILE_NAME));
+                                    ZipEntry entry = new ZipEntry((String) exchange.getIn().getHeader("zipFileName"));
```
Croway, line 357: just wondering if zipFileName is correct, where does It come from? if it comes from camel shouldn't it be CamelZipFileName or something similar?

bsmolicek: It is set by Camel core’s [ZipIterator](https://github.com/apache/camel/blob/main/components/camel-zipfile/src/main/java/org/apache/camel/dataformat/zipfile/ZipIterator.java#L143-L151) when it iterates each ZIP entry:

```
  String zipFileName = currentEntry.getName();
  answer.setHeader("zipFileName", zipFileName);
  answer.setHeader(Exchange.FILE_NAME, FileUtil.stripPath(zipFileName));
```

looks weird, but should be correct, it's also used by https://github.com/apache/camel-spring-boot/blob/main/components-starter/camel-zipfile-starter/src/test/java/org/apache/camel/dataformat/zipfile/springboot/ZipFileMultipleFilesSplitterTest.java#L97

Croway: right, it is even part of the 4.22 upgrade guide `from `zipFileName` for zip instead of `CamelFileName`.` I still wonder if this is 100% correct though, I'll investigate. But the PR is ok

PR #1945, opened 2026-09-07
CAMEL-24498: restore observability services bind default

Croway:
## Summary

- Restore the Spring Boot management listener default bind address for observability services.
- Keep the existing health-endpoint hardening.
- Add a regression test and regenerate the starter documentation.


squakez (approved):
Thanks for the quick fix!

PR #1951, opened 2026-09-09
CAMEL-24501: Only copy configured starter options onto the Camel target

Croway:
Fixes item 2 of [CAMEL-24501](https://issues.apache.org/jira/browse/CAMEL-24501), deferred from #1935.

## Problem

Every generated `*ComponentConfiguration` / `*DataFormatConfiguration` / `*LanguageConfiguration` class materialises
its catalog defaults as field initializers (`private Boolean skipRequestHeaders = false;`). The generated customizer
copied every non-null field onto the Camel target, and Camel core applies that customizer to any component, data
format or language resolved into the context, including a bean the application registered itself:

```java
@Bean("http")
HttpComponent http() {
    HttpComponent http = new HttpComponent();
    http.setSkipRequestHeaders(true); // silently reverted to the catalog default false at startup
    return http;
}
```

The runtime could not tell a configured value from an unused default, since both are plain non-null values.

## Fix

Spring Boot already records every property it binds onto a `@ConfigurationProperties` class in
`BoundConfigurationProperties` (public API since Spring Boot 2.3, used by the actuator `configprops` endpoint).
`CamelPropertiesHelper.copyConfigurationProperties` now copies an option only when that record contains the
property, or any property below it for a map or list option. Name comparison is dash and case insensitive, so
environment variables and relaxed spellings match.

Nothing generated changes. The catalog default stays as field initializer, which is what
`spring-configuration-metadata.json`, IDE completion and the generated documentation pages read the default from,
and the getters keep returning it. This is the same pattern Spring Boot uses for its own properties classes; what
differs is that Camel layers external configuration on top of user supplied beans, like Camel Main does, so the
starter needs to know what was configured, and Camel Main gets that for free by reading only the properties that
exist.

Whatever is left unbound after the copy was configured by the application, so the fail-fast and
`camel.springboot.lenient-configuration-binding` paths from #1935 apply unchanged, and the environment lookup that
#1935 added to classify an unbound option is replaced by the same record. Configuring an option to the same value
as its catalog default still takes effect, which is the objection #1935 raised against comparing values.

If the record is unavailable (no `ConfigurationPropertiesBindingPostProcessor` in the context) every option is
copied as before, with a WARN.

## Tests

- `CamelPropertiesHelperTest`: the configuration class is now bound by Spring Boot from the test properties, like a
  generated one. An unconfigured default does not overwrite the target, an option configured to its default is
  applied, a configured option that cannot be set fails, and an unconfigured one that cannot be set does not.
- `HttpComponentUserBeanTest`, `GsonDataFormatUserBeanTest`, `JsonPathLanguageUserBeanTest`: end to end on a
  component, a data format and a language registered as user beans, each with a value set in Java that differs from
  the catalog default, next to an option set through properties.

Run green: `core/camel-spring-boot` (full suite), `camel-http-starter`, `camel-gson-starter` and
`camel-jsonpath-starter`.

## Docs

`starter-configuration.adoc` now states that external configuration is applied on top of a user bean, and that a
parameter not set externally keeps the value set on the bean.

_Claude Code (Claude Fable 5.1) on behalf of Federico Mariani_

### DIFF
--- a/core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/util/CamelPropertiesHelper.java
+++ b/core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/util/CamelPropertiesHelper.java
@@ -35,9 +35,8 @@ import org.apache.camel.util.ObjectHelper;
 import org.apache.camel.util.StringHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.boot.context.properties.BoundConfigurationProperties;
 import org.springframework.boot.context.properties.source.ConfigurationPropertyName;
-import org.springframework.boot.context.properties.source.ConfigurationPropertySource;
-import org.springframework.boot.context.properties.source.ConfigurationPropertySources;
 import org.springframework.context.ApplicationContext;
 
 /**
@@ -67,20 +66,22 @@ public final class CamelPropertiesHelper {
      * Copies the options from a generated Spring Boot configuration class onto the Camel component, data format or
      * language it configures.
      * <p/>
-     * The options that belong to the auto configuration layer itself (<tt>enabled</tt> and <tt>customizer</tt>) are
-     * removed first, as they are not options on the target bean.
+     * Only the options the application configured are copied, as recorded by Spring Boot in
+     * {@link BoundConfigurationProperties} when it bound the configuration class. The other options merely carry the
+     * default value the generator took from the Camel catalog as field initializer, and must not overwrite the value
+     * the target already holds, such as one set programmatically on a user supplied bean. The options that belong to
+     * the auto configuration layer itself (<tt>enabled</tt> and <tt>customizer</tt>) are never copied, as they are not
+     * options on the target bean.
      * <p/>
-     * An option that cannot be set on the target and that the application configured explicitly fails fast with an
-     * {@link IllegalArgumentException}, instead of being dropped without a trace. An option that cannot be set and
-     * that only carries its catalog default is logged at DEBUG, as there is nothing the application can do about it
-     * and the target keeps its own default. Set {@link #LENIENT_CONFIGURATION_BINDING} to <tt>true</tt> to log an
-     * explicitly configured option at WARN and continue, instead of failing.
+     * A configured option that cannot be set on the target fails fast with an {@link IllegalArgumentException},
+     * instead of being dropped without a trace. Set {@link #LENIENT_CONFIGURATION_BINDING} to <tt>true</tt> to log it
+     * at WARN and continue, instead of failing.
      *
      * @param camelContext
      *                           the CamelContext
      * @param applicationContext
-     *                           the Spring application context, used to tell an explicitly configured option from a
-     *                           catalog default
+     *                           the Spring application context, used to tell a configured option from a catalog
+     *                           default
      * @param propertyPrefix
      *                           the configuration prefix of the source, such as <tt>camel.component.http</tt>
      * @param source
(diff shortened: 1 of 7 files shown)


core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/util/CamelPropertiesHelper.java
```diff
@@ -97,6 +104,12 @@ public static void copyConfigurationProperties(CamelContext camelContext, Applic
         Map<String, Object> properties = getNonNullProperties(camelContext, source);
         properties.keySet().removeIf(key -> AUTO_CONFIGURATION_OPTIONS.contains(key.toLowerCase(Locale.US)));
 
+        boolean tracked = source instanceof ConfiguredOptionsAware;
+        if (tracked) {
+            Set<String> configured = lowerCase(((ConfiguredOptionsAware) source).configuredOptions());
+            properties.keySet().removeIf(key -> !configured.contains(key.toLowerCase(Locale.US)));
```
davsclaus, line 110: Question, not blocking: with this filter, a catalog-only option that no longer binds on the target is dropped before `doSetCamelProperties` runs, so the DEBUG drift log below only fires for a source that is not `ConfiguredOptionsAware`, which in practice means never for generated starters. #1935 introduced that log to surface generator/catalog drift. Is losing that signal for the generated classes intended? If so it may be worth a sentence in the Javadoc so the DEBUG branch is not read as covering the generated case.

davsclaus (approved):
Thanks for this follow-up to #1935, the approach is clean and the regeneration is exactly as described.

**Verified**

- The generated diff is uniform: across the 474 regenerated classes the only added line is the `optionConfigured("...")` call in each setter. No `src/main/docs/*.json` or `.adoc` under the starters changed.
- No hand-written subclass is left untracked. The only non-generated subclasses of the three `*ConfigurationPropertiesCommon` classes are the empty core `ComponentConfigurationProperties` and siblings, and only generated auto-configurations call `copyConfigurationProperties`.
- The fix relies on Spring Boot's `JavaBeanBinder` invoking the setter for every bound property. Checked against the spring-boot 4.1.1 bytecode: a non-null bound value always goes through `setValue`, including map and list values.
- Ran locally on the PR branch: the generator plugin tests and `CamelPropertiesHelperTest` (14 tests) pass.
- Conventions: commit message follows `CAMEL-XXXX: ...`, the PR body links the Jira, the interface addition is backwards compatible, no new dependencies, and the new tests use no `Thread.sleep`.

**Non-blocking suggestions**

1. **Upgrade guide note.** This is a visible behaviour change: values set in Java on a user-registered component, data format or language now survive startup, and a value set programmatically on a configuration bean that cannot bind now fails fast instead of being logged at DEBUG. The apache/camel `camel-4x-upgrade-guide-4_23.adoc` already has a `=== camel-spring-boot` section for other starter default changes, so a short entry there would help users who relied on the old behaviour.
2. The new mojo test only asserts the `setterBody` string. That is fine given the full regeneration plus the three end-to-end starter tests, just noting it.
3. The PR is still marked as draft.

This review covers project rules and conventions only and does not replace specialized AI review tools or static analysis such as SonarCloud.

_Claude Code on behalf of Claus Ibsen_

core/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/util/CamelPropertiesHelper.java
Croway: The branch was rewritten after this review was started, so this line no longer exists. The current version does not track anything in the generated classes: `copyConfigurationProperties` keeps only the options that Spring Boot recorded as bound in `BoundConfigurationProperties`, and the DEBUG branch is gone with the environment lookup that fed it.

Losing that signal is intended. An option that only carries its catalog default is never offered to the target any more, so there is nothing to log about it; a configured option that cannot be set still fails fast (or logs at WARN when lenient). The Javadoc now states that only configured options are copied, so the method no longer reads as covering the unconfigured case.

_Claude Code (Claude Fable 5.1) on behalf of Federico Mariani_

davsclaus (approved):

PR #1994, opened 2026-09-23
camel-salesforce-starter: fix Spring Boot binding of complex component option types (Jackson 2.18+ / Spring Boot 4.1.1)

smongiar:
## Problem

Jackson 2.18+ (shipped with Spring Boot 4.1.1) changed `JacksonFeatureSet` so it is no longer bindable via Spring Boot's `Binder`. When `SalesforceComponentConfiguration` is loaded at startup, Spring Boot traverses:

```
camel.component.salesforce.object-mapper
  .serializer-provider.generator.write-capabilities
    -> JacksonFeatureSet<StreamWriteCapability>  (not bindable)

camel.component.salesforce.config.object-mapper
  .serializer-provider.generator.write-capabilities
    -> JacksonFeatureSet<StreamWriteCapability>  (not bindable)
```

causing a `BindException` at startup even when the user has not set any of those properties.

## Root cause

`SalesforceComponentConfiguration` holds a raw `ObjectMapper` instance field (and a `SalesforceEndpointConfig` field which itself contains one). Spring Boot 4.1.1 now eagerly introspects all field types at context startup — even when the properties are not configured.

## Fix

The data format and language configuration generators in `SpringBootAutoConfigurationMojo` already handle this correctly: they convert any complex-type (object / duration) field to `String` in the generated configuration class, relying on Camel's type-converter / bean-registry lookup at runtime (e.g. `#bean:myObjectMapper`).

The **component** generator (`createComponentConfigurationSource`) was missing this conversion. This PR applies the same `complex => String` rule already present in `createDataFormatConfigurationSource` and `createLanguageConfigurationSource`, bringing component configuration generation in line with the rest of the project.

## Changes

- `SpringBootAutoConfigurationMojo.java`: in the component options loop, compute `boolean complex = isComplexTypeOrDuration(option) && isBlank(option.getEnums())` and set `type = "java.lang.String"` when true — exactly mirroring the data format / language pattern
- `SalesforceObjectMapperBindingTest.java`: new test that verifies the Spring Boot context loads without a `BindException` when `camel-salesforce-starter` is on the classpath

Generated starter files (`SalesforceComponentConfiguration.java` etc.) are not included — CI regenerates them automatically after this change is merged.


Croway:
I think that a Rebase is needed, there are many unrelated commits

